PR #3, opened 2016-11-08
[FIX WEBSITE-212] - Remove Voltron text

michaelmccaskill:
Related to issue [WEBSITE-212]

Summary of this pull request: 

Remove Voltron text.

full view

zoomed

### DIFF
--- a/app/components/Main.jsx
+++ b/app/components/Main.jsx
@@ -84,7 +84,7 @@ class Main extends React.PureComponent {
           <form ref="form" action="#" id="plugin-search-form"
               className={classNames(styles.HomeHeader, { showFilter: this.props.showFilter }, 'HomeHeader jumbotron')}
               onSubmit={this.handleOnSubmit}>
-            <h1><span className="logo">project</span>Voltron</h1>
+            <h1><span className="logo">Jenkins</span>Plugins</h1>
             <p className="tagline">The strength of many, shared by all.</p>
             <nav className={classNames(styles.navbar, 'navbar')}>
               <div className="nav navbar-nav">


app/components/Main.jsx
```diff
@@ -84,7 +84,7 @@ class Main extends React.PureComponent {
           <form ref="form" action="#" id="plugin-search-form"
               className={classNames(styles.HomeHeader, { showFilter: this.props.showFilter }, 'HomeHeader jumbotron')}
               onSubmit={this.handleOnSubmit}>
-            <h1><span className="logo">project</span>Voltron</h1>
+            <h1><span className="logo">Jenkins</span>Plugins</h1>
             <p className="tagline">The strength of many, shared by all.</p>
```
daniel-beck, line 88: This line is still weird.

daniel-beck:
For design changes, could you include screenshots in PR description/comments?

app/components/Main.jsx
michaelmccaskill: Ok I'll kick it back to Gus and see options we can create.

michaelmccaskill:
Sure.

michaelmccaskill:
I added screenshots showing the current proposal but I will change them when I have an updated text.

daniel-beck:
It looks like the current text style is tailored for the old project voltron message and looks a bit weird when it's just "Jenkins Plugins", no?

michaelmccaskill:
Should we lose the image, too?

rtyler:
@michaelmccaskill Nah, the voltron Jenkins is neat and fits the plugins concept nicely. :smi

daniel-beck:
I agree with @rtyler. We've been using modified Jenkins logos a lot and they're very popular. So even if we're not calling it Voltron anymore, having a Mr Jenkinstron is fine.

michaelmccaskill:
I'm closing this until Gus has time to come up with a new text.

PR #9, opened 2016-11-10
[FIX WEBSITE-212] Replace project voltron message

daniel-beck:
Screenshot

Boring, but can always be optimized later.

Supersedes #3.

@rtyler @michaelmccaskill WDYT?


rtyler (approved):

michaelmccaskill:
If you're fine with it then I am 👍

PR #10, opened 2016-11-11
[FIX:WEBSITE-227] stylesheet corrections

gusreiber:
Related to issue # WEBSITE-227

Summary of this pull request: 
Stylesheet conflict resolutions

### DIFF
--- a/public/assets/css/base.css
+++ b/public/assets/css/base.css
@@ -1,3 +1,8 @@
+/*TODO:=====top offset for IO site should come form IO css =============*/
+section.modalview > .dialog {padding-top:3.25rem}
+/********/
+
+body .fixed.top {z-index:999}
 img {max-width:100%}
 body {overflow-x:hidden}
 body .showResults.item-finder > #plugin-search-form.HomeHeader{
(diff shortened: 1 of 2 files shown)


gusreiber:
@michaelmccaskill @daniel-beck please review
screen shot 2016-11-11 at 2 03 23 pm
screen shot 2016-11-11 at 2 02 39 pm
screen shot 2016-11-11 at 2 02 30 pm

The font icons showing as boxes is a CORS issue that would not happen in production

daniel-beck:
The screenshot shows there's still an a:visited style for the download popup when it doesn't exist on the main site.

Does this also address WEBSITE-226? Or why is that resolved?

gusreiber:
@daniel-beck 226 and 225 are separate symptoms of this same issue, so I marked them as dups. The plugin site does have a different visited link color, yes. Are you asking for it to not have that?

daniel-beck:
@gusreiber Ah, okay, I was only looking at the resolution (which is Fixed rather than Duplicate).

The idea of WEBSITE-227 is seamless integration into the Jenkins website. Sharing the top bar navigation, periodically synchronizing it to not be outdated (and no need to have the Download button removed)… but I noticed that the style on the plugins site resulted in a top bar that looks different. Hence WEBSITE-227 – styles specific to the "plugin site" should be scoped to its container element to not (AFAIU) override main site styles. (Assuming I'm guessing correctly what the cause of the differences is.)

PR #12, opened 2016-11-15
[FIX WEBSITE-234] - Add support for displaying versions

michaelmccaskill:
Related to issue [WEBSITE-234]

Summary of this pull request: 

If a file in the root directory is present with keys `plugin-site` and
`plugin-site-api` with their respective values being the short form of
the HEAD commit hash, the version partial will be injected into DOM
that's returned from jenkins.io

screen shot 2016-11-15 at 11 42 21 am

### DIFF
--- a/server.js
+++ b/server.js
@@ -35,6 +35,9 @@ app.use(jsPath, express.static('./dist/client'));
 app.engine('hbs', exphbs({extname: '.hbs'}));
 app.set('view engine', 'hbs');
 
+let pluginSiteVersion = null;
+let pluginSiteApiVersion = null;
+
 const downloadHeader = () => {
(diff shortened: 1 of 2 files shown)


michaelmccaskill:
@daniel-beck I'm interested if you visually accept this.

daniel-beck:
Seems fine. If anything, could be shortened a bit. Stack Overflow for example just says

> rev 2016.11.15.4191

So something like

> UI 99a0e31 / API 1a342b6

would also work.

(perhaps even linking to the repositories…)

But again, seems fine. Shorten if you like the idea, merge if you don't :wink:

michaelmccaskill:
@daniel-beck Ok great thanks for the feedback. I'll make these adjustments. I like the link idea as well.

PR #36, opened 2017-03-08
[FIX WEBSITE-240] - Use different point color

michaelmccaskill:
Also took it as an opportunity to upgrade the chart library as chart.js
is 2.x now.


daniel-beck:
:+1: assuming this looks similar to my "screenshot" in the issue.

michaelmccaskill:
Yeah I attached another screenshot to the ticket but I'll put it here, too.

screen shot 2017-03-08 at 5 10 01 pm

PR #53, opened 2019-07-20
[INFRA-2184] fix the trend which could just has one point

LinuxSuRen:
Summary of this pull request: make sure the trend chart will have the right points which not more than 12 or euqal with total number


LinuxSuRen:
Before:
image
After:
image

rtyler (approved):

PR #56, opened 2019-09-02
[WEBSITE-642] - Collapse implied dependencies by default

oleg-nenashev:
See https://issues.jenkins-ci.org/browse/WEBSITE-642 . Some plugins have dozens of implied dependencies, and they look really weird on the plugin site. It would be great to collapse implied dependencies by default.

## Before

image

## Now

Default:

image

Expanded:

image

CC @jglick @daniel-beck


daniel-beck:
I do not think it makes sense to hide them, as these are actual dependencies as far as admins on current releases of Jenkins are concerned. This serves as a motivation for plugin maintainers to fix their baseline to get rid of them (I currently recommend at least 2.121.x as baseline).

oleg-nenashev:
What if we explicitly add the update suggestion to the popup window? I
agree we should encourage plugin maintainers to update somehow, but I am
not sure why we want to confuse users. If a plugin is not maintained, ugly
plugins.jenkins.io layout won't help

daniel-beck:
> If a plugin is not maintained, ugly plugins.jenkins.io layout won't help

These are the answer to the question "If I use this, what dependencies will it pull in?" – no different from a plugin that explicitly declares many dependencies, like https://plugins.jenkins.io/workflow-aggregator, except it only applies to recent baselines. Since we expect users to be on up to date instances, this distinction doesn't matter.

If any would deserve to be hidden, it's optional ones, but I think we still (mostly) treat implied dependencies as real dependencies.

But just hiding these to hide the negative impact of old plugin core dependencies is not the solution.

oleg-nenashev:
I can also add a collapse for optional dependencies if it helps to move the story forward. But it really needs more feedback if we want to proceed in any direction.

CC @jglick @MarkEWaite @zbynek @halkeye and @jenkins-infra/contributors who might provide feedback.

daniel-beck:
@oleg-nenashev To clarify, I'm not saying to collapse both optional and implied. The _only_ kind of dependency that makes sense to not show like explicit mandatory dependencies is optional dependencies.

zbynek:
I think it might be better to find a solution that does not break the layout for big number of dependencies. That would benefit not only unmaintained plugins, but also important ones like
https://plugins.jenkins.io/blueocean
https://plugins.jenkins.io/workflow-aggregator

A nice way is to use Readme / Dependencies / Releases tabs, similar to NPM (e.g. https://www.npmjs.com/package/jenkins-jobs-runner ), but maybe expandable heading __Dependencies (number)__ would be good enough.

daniel-beck:
Blue Ocean is also a nice example for a maintainers list that's way too long.

halkeye:
> Blue Ocean is also a nice example for a maintainers list that's way too long.

And most of them don't work on it. Should get moved to contributors.

I'll take a look at this PR when I get a chance

halkeye:
What was the final decision on this, collapse or not collapse?

It might be easier just to split them up and have 2 headers.

timja:
I would be in favour of sub-pages / panels and collapse them all by default. They can checked if they need to, but do they really need to be there by default and take up lots of page space (especially on mobile)

oleg-nenashev:
@halkeye @timja My preference would be to introduce tabs on the top: Overview (current layout), Changelog (in my TODO list), Dependencies, Security advisories, Contributing. Or something like that. It would make the interface more extensible in the future.

jglick:
> These are the answer to the question "If I use this, what dependencies will it pull in?"

That is one possible question, but it is not obvious to me that this is the question someone actually has in mind when they look at a plugin page. I think the PR as originally conceived is fine and useful.

daniel-beck:
> but it is not obvious to me that this is the question someone actually has in mind when they look at a plugin page

What question do you think should be answered here?

And how would you address "I installed X and it pulled in more than the dependencies in the list, what's going on!?"

oleg-nenashev:
No longer relevant after the website rework

PR #61, opened 2019-09-11
Remove red `(required)` from mandatory dependencies

jglick:
It is the default, and most common, scenario for a plugin-to-plugin dependency to be required. There is no reason to highlight this in red as if it were an error the user must pay attention to. In fact we should not even be displaying **(required)** at all; it is just visual noise.


oleg-nenashev (approved):
+1, let's keep it simple

PR #71, opened 2019-12-02
Make asciidoc admonitions more noticeable

rsandell:
Slightly mimics how it looks on GitHub.

Related to issue # 
Part of https://issues.jenkins-ci.org/browse/WEBSITE-690

Summary of this pull request: 

If the plugin documentation is migrated to asciidoc and admonitions is used (the note and info boxes) it doesn't stick out as it should on the plugin site.

This is what it will look like after the change:

pluginsite_admonition_change

### DIFF
--- a/public/assets/css/base.css
+++ b/public/assets/css/base.css
@@ -901,3 +901,12 @@ a.card:hover {
     border-bottom: 1px solid #CAD0D2;
     border-radius: 4px;
 }
+
+.modalview .content .main table {
+  border: 1px solid silver;  
+}
+
+.modalview .content .main td {
+  padding: 1em;
+}
+


oleg-nenashev (approved):
Looks right

timja (approved):

timja:
Change looks fine, should it be the plugin's responsibility to add security warnings though? 
I would think it should be done through the plugin-site?

(doc in question: https://github.com/jenkinsci/gerrit-trigger-plugin/blob/master/docs/README.adoc)

zbynek:
agreed with @timja , discussion on that topic can go to https://github.com/jenkinsci/gerrit-trigger-plugin/pull/405

Does not block this PR though.

rsandell:
The security warning is just something that followed along with the wiki extractor tool, there are other info and note admonitions in the docs.

rsandell:
So who should I ping to get this merged?

daniel-beck:
> The security warning is just something that followed along with the wiki extractor tool

Right, should be removed though as the plugin site has that content in the sidebar (and otherwise was dynamically generated anyway).

timja:
@daniel-beck @oleg-nenashev and @kohsuke are the maintainers according to the github team (I've requested to join via GitHub)

oleg-nenashev (approved):
Would rather automatically remove ones which are already supported by the plugin site, but fine as improvement

PR #91, opened 2020-01-15
Switch to pulling wiki content directly since it seems to fairly often fail to pull from api

halkeye:
@olblak @timja can you pull logs from API to see why they are failing. When i hit the api endpoint it has content populated.

https://plugins.jenkins.io/github-autostatus is an example of it not working.

Added retries also cause we can take more time to build the site that api can't do


timja (commented):
If I’m reading this right it’s getting wiki content directly but GitHub content still comes from the API?

halkeye:
yep, 1 of 4

https://github.com/jenkins-infra/plugin-site-api/blob/bd2a5fe337ddf9526666ef5a7ad6a3f5bb388c6f/src/main/java/io/jenkins/plugins/services/impl/HttpClientWikiService.java#L49-L52

timja (approved):

PR #96, opened 2020-01-20
Defer external css

halkeye:
I'm working on getting the build to build locally (just takes time) so i can upload it to netlify for testing, but it made first paint a lot faster.

https://web.dev/defer-non-critical-css/ says `link rel="preload" as="style"` will defer styles, then you add a <noscript> to load if javascript isn't available

So my work around, to wait 250 seconds, then switch anything left over preloading to regular loading, seems to be a good backup/work around.

Works fine in edge/chrome/notfirefox (https://caniuse.com/#feat=link-rel-preload) - https://jenkins-plugins.g4v.dev/localization-zh-cn


halkeye:
image
https://jenkins-plugins.g4v.dev/localization-zh-cn

halkeye:
01:30 <halkeye> woo, was able to get a build now that I figured out it was essentially setting up 1700 requests at once, oops :D
01:30 <halkeye> https://github.com/jenkins-infra/plugin-site/pull/96#
01:30 <halkeye> first time to paint is super super fast
01:32 <halkeye> does not work in firefox :(
01:35 <halkeye> :( https://support.mozilla.org/en-US/questions/1240728

halkeye:
So my work around, to wait 250 seconds, then switch anything left over preloading to regular loading, seems to be a good backup/work around.

Works fine in edge/chrome/notfirefox (https://caniuse.com/#feat=link-rel-preload) - https://jenkins-plugins.g4v.dev/localization-zh-cn

PR #103, opened 2020-01-23
Fix for #102 where the warning icon is missing

### LINKED ISSUE #102: Plugin site no longer shows active security warnings
Copied from https://issues.jenkins-ci.org/browse/INFRA-2425

https://plugins.jenkins.io/websphere-deployer is a plugin that should show a warning exclamation icon next to the plugin name, that opens a popup listing the active security warnings.

(The warnings in the description text are probably just legacy content from the wiki.)

Additionally, warning indicators appear to have been list on the plugins list (search results etc.) as well, although I'm less sure they existed before.

"Previous security warnings" in the sidebar (e.g. https://plugins.jenkins.io/git ) still exist, so it's unlikely to be a data problem.

cc @daniel-beck / @halkeye

halkeye:
Related to issue 
Fixes #102

Summary of this pull request:  Restores the badge box being visible

image

### DIFF
--- a/src/styles/base.css
+++ b/src/styles/base.css
@@ -234,11 +234,8 @@ transform: scale(-1, 1);
   margin-right: 15px;
 }
 #pluginPage #grid-box .download {margin:2rem 0 2rem -.5rem;}
-#pluginPage #grid-box > .content {padding:0 !important; max-height:100% !important; overflow-x: hidden;}
-#pluginPage #grid-box > .content .title {margin-bottom:2rem; vertical-align:top}
-#pluginPage #grid-box > .content .content {min-height:25rem}
-#pluginPage #grid-box > .content .badge-box {display: inline-block;}
-#pluginPage #grid-box > .content .badge-box  > .badge {
+#pluginPage #grid-box .badge-box {display: inline-block;}
+#pluginPage #grid-box .badge-box  > .badge {
   background: no-repeat top;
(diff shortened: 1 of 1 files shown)


timja (approved):
🤦‍♂ I thought there was a bigger issue, tested locally.
Modal isn't that nice, but same as before I assume

timja:
Hmm, it's missing a pointer / hand click to show it's clickable, was it like that before?

(tbh I don't think it's obvious this is clickable anyway and I don't like the UX of this feature)

halkeye:
>  Modal isn't that nice, but same as before I assume

i think the modal used to be bigger. I switched it to bootstrap modal from the home brewed one

> Hmm, it's missing a pointer / hand click to show it's clickable, was it like that before?

As far as I can tell it was like that. I can add `cursor: pointer` easy. Agreed UX sucks

halkeye:
Gonna merge, we can fix css later

daniel-beck:
> i think the modal used to be bigger. I switched it to bootstrap modal from the home brewed one

Yeah, filled the entire screen, was pretty horrible for how little content it contained.

This one is much nicer, but perhaps the titles could be a bit smaller? Seems excessive.

halkeye:
@daniel-beck I can certainly make it wider, but it might be worth getting UX guild's opinion (and a new issue)

PR #154, opened 2020-03-10
importing the sentry ignore list from sentry website

halkeye:
Not entirely sure how to test this, but importing the list from the sentry website

https://docs.sentry.io/platforms/javascript/#decluttering-sentry

cc @ndmanvar

### DIFF
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -116,11 +116,49 @@ module.exports = {
                 dsn: 'https://495b0bd32a5e4a2287c3fe4b061ee24f@sentry.io/1882460',
                 environment: process.env.NODE_ENV,
                 enabled: ['production', 'stage'].includes(process.env.NODE_ENV),
-                blacklistUrls: [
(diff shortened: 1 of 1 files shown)


timja (approved):

oleg-nenashev (approved):

ndmanvar:
@halkeye - https://github.com/getsentry/sentry-javascript/blob/master/packages/core/test/lib/integrations/inboundfilters.test.ts is how we test some of this (disclaimer: didn't take an extremely in-depth look. just sharing in case it will help)

halkeye:
> @halkeye - https://github.com/getsentry/sentry-javascript/blob/master/packages/core/test/lib/integrations/inboundfilters.test.ts is how we test some of this (disclaimer: didn't take an extremely in-depth look. just sharing in case it will help)

Those look like unit tests, so won't really help confirm they work with gatsby plugin

Every local thing I tried says its good, so YOLO!

PR #199, opened 2020-04-20
Remove Jira table, improve side panel links

zbynek:
3 minor SEO related fixes
* remove the Jira confluence macro which doesn't work anywhere and breaks usability on mobile
* increase spacing for links in side panel (reported by Chrome audit)
* meaningful description for the guidelines link (also from Chrome audit)

### DIFF
--- a/plugins/gatsby-source-jenkinsplugins/utils.js
+++ b/plugins/gatsby-source-jenkinsplugins/utils.js
@@ -31,6 +31,8 @@ function getContentFromConfluencePage(url, content) {
 
     // Remove any table of contents
     $('.toc').remove();
+    // remove jira issue table
+    $('.jira-table.conf-macro.output-block').remove();
 
     // Replace href/src with the wiki url
     $('[href]').each((idx, elm) => {
--- a/src/templates/plugin.jsx
+++ b/src/templates/plugin.jsx
@@ -83,8 +83,8 @@ function PluginPage({data: {jenkinsPlugin: plugin}}) {
                     </div>
                 
                     <h5>Links</h5>
-                    {plugin.scm && plugin.scm.link && <div><a href={plugin.scm.link}>GitHub</a></div>}
-                    <div><a href={`https://javadoc.jenkins.io/plugin/${plugin.name}`}>Javadoc</a></div>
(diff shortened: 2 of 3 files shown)


src/templates/plugin.jsx
```diff
@@ -96,8 +96,8 @@ function PluginPage({data: {jenkinsPlugin: plugin}}) {
                             <a href={plugin.wiki.url} target="_wiki">Jenkins Wiki</a>
                             {' the '}
                             <a href="https://groups.google.com/forum/#!msg/jenkinsci-dev/lNmas8aBRrI/eL3u7A6qBwAJ" rel="noopener noreferrer" target="_blank">read-only state</a>
-                            {'. We recommend moving the plugin documentation to GitHub, see the guidelines '}
-                            <a href="https://jenkins.io/blog/2019/10/21/plugin-docs-on-github/" rel="noopener noreferrer" target="_blank">here</a>
```
halkeye, line 100: so 👍, "here" is such bad practice.

plugins/gatsby-source-jenkinsplugins/utils.js
```diff
@@ -31,6 +31,8 @@ function getContentFromConfluencePage(url, content) {
 
     // Remove any table of contents
     $('.toc').remove();
+    // remove jira issue table
+    $('.jira-table.conf-macro.output-block').remove();
```
halkeye, line 35: why not just .jira-table?

halkeye (approved):

plugins/gatsby-source-jenkinsplugins/utils.js
zbynek: Could anyone use `jira-table` in some other context where it makes sense to keep it? I wasn't 100% sure.

halkeye: yea, more specific is good since its auto generated anyways.

halkeye: I'm not sure we can easily do it on plugins-api, but since all wiki fetching is done by plugin-site for now, i'm happy with this.

Issue #138, opened 2020-02-27
Add a "Report a problem" button to the footer


oleg-nenashev:
it would be great to add a new "report a problem" button so that Jenkins users could quickly create GitHub issues. It would be similar to what @timja did in https://github.com/jenkins-infra/jenkins.io/pull/2896 for jenkins.io

apavlidi:
Hello @oleg-nenashev 

I tried to add it, but couldn't find the file that generates the **\<footer\>** tag. Is it even on this project or it pulled from an external source? 

Thanks :call_me_hand:

halkeye:
so https://github.com/jenkins-infra/plugin-site/blob/master/utils.js generates/copies all the jenkins.io html to the plugin site. I wouldn't try to modify that at all.

https://github.com/jenkins-infra/plugin-site/blob/master/src/templates/plugin.jsx is the specific plugin page, and probably is the best place to start.

GuelorEmanuel:
@oleg-nenashev I think I have a quick solution for this. Can you assign this to me, as long as it's ok with @apavlidi?

oleg-nenashev:
Thanks for your contribution @GuelorEmanuel !

PR #237, opened 2020-05-22
Add a "Report a problem" button to the footer

GuelorEmanuel:
#138

### DIFF
--- a/src/components/ReportAProblem.jsx
+++ b/src/components/ReportAProblem.jsx
@@ -2,9 +2,11 @@ import React from 'react';
 import PropTypes from 'prop-types';
 
 function ReportAProblem({reportProblemTitle, reportProblemUrl, reportProblemRelativeSourcePath}) {
+    const title = `Report a problem with ${reportProblemRelativeSourcePath}`;
+    
     return (
         <p className="box">
-            <a href={`https://github.com/jenkins-infra/plugin-site/issues/new?labels=bug&template=4-bug.md&title=${reportProblemTitle} page - TODO: Put a summary here&body=Problem with the [${reportProblemTitle}](https://plugins.jenkins.io${reportProblemUrl}) page, [source file](https://github.com/jenkins-infra/plugin-site/tree/master/src${reportProblemRelativeSourcePath})%0A%0ATODO: Describe the expected and actual behavior here %0A%0A%23%23 Screenshots %0A%0A TODO: Add screenshots if possible %0A%0A%23%23 Possible Solution %0A%0A%3C!-- If you have suggestions on a fix for the bug, please describe it here. --%3E %0A%0AN/A`} title={reportProblemTitle}>
+            <a href={`https://github.com/jenkins-infra/plugin-site/issues/new?labels=bug&template=4-bug.md&title=${reportProblemTitle} page - TODO: Put a summary here&body=Problem with the [${reportProblemTitle}](https://plugins.jenkins.io${reportProblemUrl}) page, [source file](https://github.com/jenkins-infra/plugin-site/tree/master/src${reportProblemRelativeSourcePath})%0A%0ATODO: Describe the expected and actual behavior here %0A%0A%23%23 Screenshots %0A%0A TODO: Add screenshots if possible %0A%0A%23%23 Possible Solution %0A%0A%3C!-- If you have suggestions on a fix for the bug, please describe it here. --%3E %0A%0AN/A`} title={title}>
                 Report a problem
             </a>
         </p>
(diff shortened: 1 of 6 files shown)


src/templates/search.jsx
```diff
@@ -15,6 +15,8 @@ import Filters from '../components/Filters';
 import ActiveFilters from '../components/ActiveFilters';
 import fetch from 'isomorphic-fetch';
 
+const path = require('path');
```
halkeye: So this will make a path polyfill be added to the front-end bundle. And I'm really not sure what the path is doing. Is it just for issue template? I don't think it needs to be anything but a simple string.

halkeye (commented):
@oleg-nenashev should weigh as the original requester, but I don't think it needs to be added to every page, then you can just report `window.location.href` and not worry about all the other items that will make `<Layout` more complicated than it is. and `document.querySelector('title').textContent` to get the title

But that would make it a runtime thing, instead of part of the static html... so maybe not.

utils.js
```diff
@@ -88,9 +88,35 @@ async function makeReactLayout() {
...
+        while (tempNode != null) {
+            if (tempNode.name === tagName) {
+                return true;
+            }
+            tempNode = tempNode.parent;
+        }
+        return false;
+    };
+
+    const handleFooterNodeAndDescendant = (node, prefix) => {
```
halkeye: Can you make a new component, i'm sad how big and complicated I had to make this function.

see the `$('#creativecommons').append('<SiteVersion />');` above
Something like `$('#.license-box').append('<ReportAProblem />');` (or even after().append() type thing)

src/templates/search.jsx
GuelorEmanuel: That's true, I am not too sure what I was thinking when I added this. Having it as a simple string does make sense. Thanks!

utils.js
GuelorEmanuel: Interesting, great suggestion. I will refactor this by creating a new component. Thanks @halkeye

utils.js
```diff
@@ -114,24 +116,29 @@ async function makeReactLayout() {
         } else if (node.type === 'comment') {
             return;
         } else if (node.type === 'text') {
-            const text = node.data;
+            const text = node.data;
```
halkeye, line 119: Not a big deal here, and I'll update eslintrc to prevent end of line space, but in general watch out for PRs with unrelated changed lines.

src/components/ReportAProblem.jsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+function ReportAProblem({reportProblemTitle, reportProblemUrl, reportProblemRelativeSourcePath}) {
+    return (
+        <p className="box">
+            <a href={`https://github.com/jenkins-infra/plugin-site/issues/new?labels=bug&template=4-bug.md&title=${reportProblemTitle} page - TODO: Put a summary here&body=Problem with the [${reportProblemTitle}](https://plugins.jenkins.io${reportProblemUrl}) page, [source file](https://github.com/jenkins-infra/plugin-site/tree/master/src${reportProblemRelativeSourcePath})%0A%0ATODO: Describe the expected and actual behavior here %0A%0A%23%23 Screenshots %0A%0A TODO: Add screenshots if possible %0A%0A%23%23 Possible Solution %0A%0A%3C!-- If you have suggestions on a fix for the bug, please describe it here. --%3E %0A%0AN/A`} title={reportProblemTitle}>
```
halkeye: i don't think the title attribute is right here, i'll leave it up to you though

https://html.com/attributes/a-title/

I'm pretty sure when you hover over this link it'll say something like "Plugin page not found" which I don't think is right for the tooltip, I think the default of "Report a problem" makes more sense.

halkeye (approved):
Thank you for this.

Overall I'm pretty happy with it. I'm going to think about if there's a gatsby way to do the <layout stuff without needing the 4 parameters everywhere, but at the moment I can't think of it.

I'll wait to see if anyone else has comments, but LGTM

timja (approved):
Nice, my only worry is that this will get reports of issues with plugin docs, rather than the plugin site itself,

It’s worth trialing anyway to see what we get, possibly routing issues to maintainers is good through here, and easy to turn off if we want

src/components/ReportAProblem.jsx
GuelorEmanuel: I'll update this in a bit.

halkeye:
WFM merging

oleg-nenashev (commented):
Thanks for the patch!

PR #238, opened 2020-05-24
Add eslint rule to strip out trailing spaces at the end of line

halkeye:
There's just been more than a few PRs that had inconsistencies

So add the rule, then ran eslint --fix


timja (approved):

oleg-nenashev (approved):
Ok?

PR #250, opened 2020-06-01
Add Releases and Issues Tabs to plugin site.

halkeye:
Related of https://github.com/jenkins-infra/plugin-site-api/pull/84 

I agree that we should add a sidebar like npm does with stats (as per @zbynek)'s suggestion. But I rather something than nothing.

Test link: https://jenkins-plugins.g4v.dev/configuration-as-code/#releases


src/components/PluginReleases.jsx
```diff
@@ -0,0 +1,55 @@
...
+    }, []);
+
+    if (isLoading) {
+        return (<div className="spinner-border" role="status">
+            <span className="sr-only">Loading...</span>
+        </div>);
+    }
+
+    return (
+        <div id="pluginReleases--container" className="container">
```
timja, line 29: This seems to be really slow at least with plugins that have lots of releases, maybe we should paginate it?

halkeye: https://jenkins-plugins.g4v.dev/git/#releases

github only returns a handful by default. I think a new UI should be done before we worry about more than that at once.

zbynek:
The tabs are a nice improvement, but why the three column layout? Now if you want to read the changes in chronological ordder and the cards are too high to fit the screen, you have to scroll up and down....

halkeye:
> The tabs are a nice improvement, but why the three column layout? Now if you want to read the changes in chronological ordder and the cards are too high to fit the screen, you have to scroll up and down....

So short answer. First attempt i only had like one or two items and i thought it was a waste of white space, but the more i used the cards the more I disliked it. 

I'll see what other options.

halkeye:
Updated the plugin site.
https://jenkins-plugins.g4v.dev/configuration-as-code/#releases

Markdown is now rendered on the server instead of a react component so we get all kinds of github goodies for the markdown

package.json
```diff
@@ -54,7 +54,7 @@
   "bugs": {
     "url": "https://issues.jenkins-ci.org/secure/RapidBoard.jspa?rapidView=1&projectKey=WEBSITE&view=detail"
   },
-  "homepage": "http://plugins.jenkins-ci.org/",
+  "homepage": "https://plugins.jenkins-ci.org/",
```
timja:
```suggestion
  "homepage": "https://plugins.jenkins.io/",
```

src/components/PluginIssues.jsx
```diff
@@ -0,0 +1,70 @@
...
+                    <caption>List of issues</caption>
+                    <thead>
+                        <tr>
+                            <th scope="col">Key</th>
+                            <th scope="col">Summary</th>
+                            <th scope="col">Assignee</th>
+                            <th scope="col">Reporter</th>
+                            <th scope="col">Priority</th>
+                            <th scope="col">Status</th>
+                            <th scope="col">Resolution</th>
```
timja, line 39: These columns are quite specific to jira, is it worth abstracting this into a `JiraPluginIssues`?

timja (approved):

src/components/PluginIssues.jsx
halkeye: Phase 2 yes, though the abstraction would be on API not ui

timja: I would assume UI would need to handle it and delegate to the correct UI though? but yes later sounds good

jetersen:
@halkeye You absolute legend!

One minor thing.
As release drafter defaults to `##` or `h2` in markdown for categories perhaps we could scale down the size of H2 on the release page?

halkeye:
> As release drafter defaults to `##` or `h2` in markdown for categories perhaps we could scale down the size of H2 on the release page?

at this point I've been working on this for a couple months. I'd really like to land it then someone with more design skills do a pass.

It is wrapped in an ID for easy tweaking in the future - https://github.com/jenkins-infra/plugin-site/pull/250/files#diff-2580831f58e49705912f31c226b98572

oleg-nenashev:
Would be great to increase width of the changelogs. Maybe we should have a single-column layout for them, similarly to GitHub releases

halkeye:
> Would be great to increase width of the changelogs. Maybe we should have a single-column layout for them, similarly to GitHub releases

it is a single column, though not full width on larger monitors.
image

oleg-nenashev:
thanks!

Issue #306, opened 2020-07-13
Move the plugin name above the tabs && show it for all tabs


oleg-nenashev:
Currently the plugin name is shown only on the documentation tab.
I propose to move it up above the tabs and to display it for all tabs.

Current plugin version can be moved to a separate row within the documentation or kept as is now

![image](https://user-images.githubusercontent.com/3000480/87308984-4a07c200-c51c-11ea-80ec-dbad6ad6eb56.png)

sachinmukherjee:
@oleg-nenashev Can I take this issue?

oleg-nenashev:
Yes!

sachinmukherjee:
Hi @oleg-nenashev  Where can I find this page? I have built this project and started the dev server using the command yarn dev. But I can see totally different page 
![image](https://user-images.githubusercontent.com/19318218/92989128-69f13d00-f4ef-11ea-9376-c01b57175d83.png)

oleg-nenashev:
Please open any plugin page

sachinmukherjee:
I have made the changes according to the issue description and it looks like this:
![image](https://user-images.githubusercontent.com/19318218/93000278-f9750b00-f544-11ea-9dc2-0a02f4a6eb51.png)
Few questions:

- Should I remove the plugin name resides below the documentation tab.

- Should I change the plugin title heading(currently h1) located above all the tabs to something else

halkeye:
> Should I remove the plugin name resides below the documentation tab.

Yes once globally is fine and looks way better

> Should I change the plugin title heading(currently h1) located above all the tabs to something else

Like smaller text or anything? I think it's fine as is

oleg-nenashev:
+1 to what @halkeye says.

halkeye:
This was handled by #363

PR #363, opened 2020-09-19
Moved the plugin name above all the tabs and made it common for all tabs

sachinmukherjee:
Related to issue https://github.com/jenkins-infra/plugin-site/issues/306 

Summary of this pull request: Moved the plugin name above all tabs and made it common for all tabs and also made version in a separate row.
image

@oleg-nenashev Please review this PR

### DIFF
--- a/src/templates/plugin.jsx
+++ b/src/templates/plugin.jsx
@@ -51,6 +51,9 @@ function PluginPage({data: {jenkinsPlugin: plugin}}) {
 
             <div className="row flex">
                 <div className="col-md-9 main">
+                    <h1 className="title">
+                        {cleanTitle(plugin.title)}
+                    </h1>
                     <ul className="nav nav-tabs">
                         {tabs.map(tab => (
                             <li className="nav-item" key={tab.id}>
@@ -60,10 +63,12 @@ function PluginPage({data: {jenkinsPlugin: plugin}}) {
                     </ul>
                     <div className="padded">
                         {state.selectedTab === 'documentation' && (<>
-                            <h1 className="title">
-                                {cleanTitle(plugin.title)}
+                            <h1 className="title-details">
                                 <PluginActiveWarnings securityWarnings={plugin.securityWarnings} />
-                                <span className="v">{plugin.version}</span>
+                                <span className="version">
+                                    {'Version: '}
+                                    {plugin.version}
+                                </span>
                                 <span className="sub">
                                     {'Minimum Jenkins requirement: '}
                                     {plugin.requiredCore}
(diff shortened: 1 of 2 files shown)


src/templates/plugin.jsx
```diff
@@ -60,10 +63,12 @@ function PluginPage({data: {jenkinsPlugin: plugin}}) {
                     </ul>
                     <div className="padded">
                         {state.selectedTab === 'documentation' && (<>
-                            <h1 className="title">
-                                {cleanTitle(plugin.title)}
+                            <h1 className="title-details">
```
timja: There should only be one h1 on the page, this section doesn’t appear to need to be a heading at all.

Can you add a screenshot of a plugin with active security warnings too?

zbynek: Changing h1 into something more suitable would also help remove all this clutter from the heading in Google search results
https://www.google.com/search?q=jenkins+git+plugin&oq=jenkins+git+plugin

sachinmukherjee: Hi @timja, I tried to find some plugin with a security warning but didn't succeed. Can you help me out with some plugin names that have some security warning?

zbynek: @sachinmukherjee please check this one: https://plugins.jenkins.io/custom-job-icon/

sachinmukherjee: @zbynek I tried the custom job icon plugin but it is not showing any security warning in my development env.
image

zbynek: you may need to specify which plugin page to build when starting the dev server:
```
export GET_CONTENT_SINGLE='custom-job-icon';yarn dev
```
image

sachinmukherjee: Do I need to install some specific tool for this because when I ran `export GET_CONTENT_SINGLE='custom-job-icon';yarn dev` it gave me an error like `'export' is not recognized as an internal or external command,
operable program or batch file.`. I'm using windows 10

halkeye: export is a linux command
you can use set instead in windows

sachinmukherjee: After following the above steps security warning is still not coming up

zbynek: @timja the screenshot of a plugin with a warning is above -- do you think it's OK for the scope of this PR or should something be done here? Personally I think it's weird that current warnings are hidden (you need to click the triangle) and warnings for old versions are shown directly. It might make sense to kill the triangle entirely and show the current warning as an [alert](https://getbootstrap.com/docs/4.0/components/alerts/) similar to the adoption message. Maybe out of scope here?

timja: the above looks ok I think, (in comparison to the previous)

yes it's very weird and needs a re-design

sachinmukherjee: Can I commit my files with the above mentioned changes?

timja: yes please

timja (approved):

oleg-nenashev (approved):
LGTM. Thanks @sachinmukherjee !

zbynek (approved):

PR #379, opened 2020-10-04
Show dependent pulgins in dependencies tab

zbynek:
Summary of this pull request: show dependent plugins (reverse dependencies) in the dependencies tab.
This is a part of my Hacktoberfest contribution.

image

### DIFF
--- a/src/components/PluginDependencies.jsx
+++ b/src/components/PluginDependencies.jsx
@@ -29,7 +29,7 @@ function PluginDependencies({dependencies, reverseDependencies} ) {
     };
     const reverseDependencyLink = (dependency) => {
         return (
-            <div key={`${dependency.node.id}-revese`}>
+             <div key={dependency.node.id`}>
                 <Link to={`/${dependency.node.id}/`}>
                     {dependency.node.title}
                 </Link>
--- a/src/components/PluginDependencies.jsx
+++ b/src/components/PluginDependencies.jsx
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 import {Link} from 'gatsby';
 import {Modal, ModalHeader, ModalBody} from 'reactstrap';
 
-function PluginDependencies({dependencies} ) {
+function PluginDependencies({dependencies, reverseDependencies} ) {
     const [isShowImplied, setShowImplied] = React.useState(false);
     const toggleShowImplied = (e) => {
         e && e.preventDefault();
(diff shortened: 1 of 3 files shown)


markjacksonfishing (approved):

timja (approved):

src/components/PluginDependencies.jsx
```diff
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 import {Link} from 'gatsby';
 import {Modal, ModalHeader, ModalBody} from 'reactstrap';
 
-function PluginDependencies({dependencies} ) {
+function PluginDependencies({dependencies, reverseDependencies} ) {
```
halkeye, line 6: reverse dependencies are things that depend on it?

src/components/PluginDependencies.jsx
```diff
@@ -27,9 +27,18 @@ function PluginDependencies({dependencies} ) {
             </div>
         );
     };
+    const reverseDependencyLink = (dependency) => {
+        return (
+            <div key={`${dependency.node.id}-revese`}>
```
halkeye: revese?

```suggestion
             <div key={`${dependency.node.id}-reverse`}>
```

or key only needs to be unique in the array its in so you could even just use id

```suggestion
             <div key={dependency.node.id`}>
```

src/templates/plugin.jsx
```diff
@@ -213,6 +221,26 @@ export const pageQuery = graphql`
     jenkinsPlugin(name: {eq: $name}) {
       ...JenkinsPluginFragment
     }
+
+    reverseDependencies: allJenkinsPlugin(
```
halkeye, line 225: niiice, no new code elsewhere, just looking at the data we already have, thats awesome.

halkeye (approved):
Minor typo but looks good to me

halkeye:
I made sure you have write access, so merge when your happy

src/components/PluginDependencies.jsx
zbynek: Yes, I borrrowed the term from [Ruby Gems](https://rubygems.org/gems/parallel/reverse_dependencies) but in the UI it's shown as "dependent plugins" because that may make more sense to the users.

oleg-nenashev:
This is superhelpful for developers. Thanks a lot @zbynek !

PR #385, opened 2020-10-07
Replace react-time-ago with react-timeago

zbynek:
Since [react-time-ago](https://www.npmjs.com/package/react-time-ago) package [can't be updated](https://github.com/jenkins-infra/plugin-site/pull/355) to latest version without changes + it seems to be responsible for some console errors in development mode, why not replace it with 
[react-timeago](https://www.npmjs.com/package/react-timeago) which seems 10x more popular and brings in fewer transitive dependencies. There is also [timeago-react](https://www.npmjs.com/package/timeago-react), but I picked the package with more downloads.

CC @halkeye

### DIFF
--- a/src/templates/plugin.jsx
+++ b/src/templates/plugin.jsx
@@ -204,14 +204,16 @@ PluginPage.propTypes = {
             }).isRequired,
             version: PropTypes.string
         }).isRequired,
-        reverseDependencies: PropTypes.arrayOf(
-            PropTypes.shape({
-                node: PropTypes.shape({
-                    id: PropTypes.string.isRequired,
-                    title: PropTypes.string.isRequired,
+        reverseDependencies: PropTypes.shape({
+            edges: PropTypes.arrayOf(
+                PropTypes.shape({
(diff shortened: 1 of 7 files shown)


src/templates/plugin.jsx
```diff
@@ -204,14 +204,16 @@ PluginPage.propTypes = {
             }).isRequired,
             version: PropTypes.string
         }).isRequired,
-        reverseDependencies: PropTypes.arrayOf(
-            PropTypes.shape({
-                node: PropTypes.shape({
-                    id: PropTypes.string.isRequired,
-                    title: PropTypes.string.isRequired,
+        reverseDependencies: PropTypes.shape({
```
zbynek, line 207: This was causing some console errors and didn't seem worth an own PR.

timja (approved):

halkeye:
lgtm

kasunsewwandika (commented):
Great thanks

kasunsewwandika (commented):
Great

PR #386, opened 2020-10-09
Move alerts to top and update their styles

zbynek:
This PR tries to address some issues with the alerts:
* Current security warnings more hidden than previous version warnings – fixed by showing security warnings in full
* Minor design issues with the alert styles – fixed by using bootstrap standard classes
* Warnings related to invalid table markup – fixed by not using tables for layout

image

</details>

<details><summary>Multiple vulnerabilities</summary>

image

</details>

<details><summary>Deprecation notice</summary>

image

</details>
CC @daniel-beck


halkeye:
> Minor design issues with the alert styles -- fixed by using bootstrap standard classes

<3 so much <3 using the standard items

> Warnings related to invalid table markup -- fixed by not using tables for layout

same

src/components/PluginGovernanceStatus.jsx
```diff
@@ -8,38 +8,25 @@ function PluginGovernanceStatus({plugin}) {
...
-                                <a href="https://jenkins.io/doc/developer/plugin-governance/deprecating-or-removing-plugin">here</a>
-                                {'.'}
-
-                            </td>
-                        </tr>
-                    </table>
+                <div className="alert alert-warning alert-with-icon">
+                    <span className="alert-icon warning" aria-label="Warning"/>
+                    <b>Deprecated:</b>
+                    {'This plugin has been marked as deprecated. '}
```
halkeye: FYI, you can use a backtick and make this one large blob, like java's 3 quotes. At some point I want to find out why eslint messes with strings

(I know you didn't make these strings just reformatted, but wanted to share)

src/styles/base.css
```diff
@@ -238,27 +237,29 @@ transform: scale(-1, 1);
...
-  vertical-align: middle;
-  text-indent: 99rem;
-  white-space: nowrap;
-  margin: 0 -.5rem 0 .5rem;
+  position: absolute;
+  top: calc(50% - 1.5rem);
+  left: 1rem;
 }
-#pluginPage #grid-box .badge-box .badge.warning {
+.security-warning-list {
```
halkeye: is there a reason this isn't scoped? is it used in multiple places?

daniel-beck:
I like this a lot. Thanks!

Could you check to make sure this looks reasonable even when a plugin has multiple active security warnings?

Two warnings can happen fairly easily and there are several plugins that have that, like https://plugins.jenkins.io/perfecto/ or https://plugins.jenkins.io/elastest/

More than that is less relevant (so doesn't need to look all that reasonable), but just for giggles could you take a screenshot of https://plugins.jenkins.io/websphere-deployer/ ?

----

I wonder whether a different phrasing for the affected versions would make sense.

Currently, if a warning is defined for an issue that has not been resolved, we still declare what the latest version is (future releases are undetermined). That could be misleading to users — any active security warning should be considered to affect the current release, even if this metadata doesn't like it.

For example, the site could detect whether the regex matches anything, and if so, word the message differently.

Perhaps not immediately related to this PR, but given the additional prominence of the version information, especially being shown at the same time as the latest version on the page, this seems different from before.

halkeye:
image

one of the many reasons I'd love to see this hooked up to netlify is that it would build demo sites per PR. I wonder how hard it would be to do that with azure sites.

timja:
This is Azure's service for it I believe: https://azure.microsoft.com/en-gb/services/app-service/static

zbynek:
@daniel-beck I suggest to remove the intervals altogether: currently they either say "afffects some versions" or "Affects version X and earlier" where X is the current version number. There is no actual difference between those two statements, it's just a bit confusing. Please see the updated screenshots in the initial comment and let me know if the wording is OK.

halkeye:
@zbynek is this still waiting on something? I say merge and we can always follow up with more changes, it looks really good as is.

daniel-beck:
@zbynek It looks good but may be slightly misleading if newer releases have been published since (for which it's essentially undefined, sometimes it's fixed, sometimes not). Overall I think this is a reasonable solution.

FWIW "some versions" has basically been retired as we now always specify the most recent known affected version so that never occurs anymore.

timja:
> @zbynek It looks good but may be slightly misleading if newer releases have been published since (for which it's essentially undefined, sometimes it's fixed, sometimes not). Overall I think this is a reasonable solution.

update center data should be updated then to remove the warning if fixed ^.^.
Users shouldn't have to figure out if it's fixed or not ideally

timja (approved):

zbynek:
>  Overall I think this is a reasonable solution.

@daniel-beck OK to merge as is?

markjacksonfishing (approved):

Issue #129, opened 2020-02-17
Optimize scaling of plugin installation trends


oleg-nenashev:
In some cases trends look and feel can be improved, especially we could chose minimum and maximum Y axis values. See the trend for Ownership plugin, for example

![image](https://user-images.githubusercontent.com/3000480/74663952-11e29a00-519d-11ea-9d15-f4f4e9d2556d.png)

ARUNMOHANRAJ471:
Hi @oleg-nenashev , I would like to take up this issue. Could you pls assign this to me.

halkeye:
Its yours now

zbynek:
@oleg-nenashev @ARUNMOHANRAJ471 would it make sense to convert the graph into a sparkline? (so basically remove the labels and make it smaller)
https://whawker.github.io/react-jsx-highcharts/examples/Sparkline/index.html has some nice demo of sparklines using a different NPM package. Together with the total number of installs this should convey the same information without taking up so much space.

zbynek:
Fixed in #402 and #445 , further improvements in progress in #447.

PR #402, opened 2020-10-16
#129 - Optimize scaling of plugin installation trends

ARUNMOHANRAJ471:
Related to issue #129 

Summary of this pull request: 
I have added an logic to dynamically set the y-axis values to improve the look and feel of the graphs which can now helps us visualize the installation trends better.

**gerrit-verify-status-reporter Plugin:**
Before:
image

After:
image

### DIFF
--- a/src/components/LineChart.jsx
+++ b/src/components/LineChart.jsx
@@ -13,21 +13,13 @@ const calculateHeight = (total) => {
 };
 
 const calculateMinMax = (data) => {
-    const lastValue = data[data.length-1];
-    let maxValue = undefined;
-    let minValue = undefined;
-    if (lastValue < 100) maxValue = 250;
-    else if (lastValue < 250) maxValue = 500;
-    else if (lastValue < 500) maxValue = 1000;
-    else if (lastValue < 1000) maxValue = 2000;
-    else if (lastValue < 2500) maxValue = 5000;
-    else if (lastValue < 7500) maxValue = 10000;
-
-    if (lastValue > 5000) minValue = 0;
-
+    const minValue = Math.min(...data);
+    const maxValue = Math.max(...data);
+    // calculate a dynamic value to center the graph
+    const scaleDifference = Math.pow(10, maxValue.toString().length-1);
     return {
-        min: minValue,
-        max: maxValue
+        min: parseInt(minValue/scaleDifference)*scaleDifference,
+        max: parseInt((maxValue/scaleDifference)+1)*scaleDifference // plus 1 to add more space in the top
     };
 };
 


halkeye:
My concern is it can mislead how many installs are being represented. A graph showing data should always have the same scale as another graph showing the exact same data.

But I'll leave it for @oleg-nenashev and @zbynek

zbynek (changes requested):
I think we should keep min value as 0 for all plugins. Otherwise this will look like big changes are happening for plugins with nearly constant number of installs.

@halkeye the max value was different for each plugin already, this just makes it more predictable.

ARUNMOHANRAJ471:
Hi @zbynek ,@halkeye If we have the min value as 0 for all the plugins the graph will look the same as like in production. If the objective of the chart is to display plugin installation trends to user, having a min of 0 in y axis will not help the user to get the insight as the changes will be mildly visible.

In this case, as suggested by @zbynek in the issue #129 , we can display the trends as a sparkline chart and can show number of weekly/monthly downloads like in npm.

Pls share your views.

halkeye:
I still don't think the graph solves anything if the scale changes each time its visible.

~Can you mock up what the sparkline would look like?~ I just went and looked at the npm version. I like it because its clear whats going on, but it has the same issue

1-5 downloads a week
image

700k downloads a week
image

They don't look different. So what is the goal of the graph? its been there longer than I have been involved. The NPM version works great if the goal is to see if anything drastic changes in the number of downloads. If thats the goal, then The current implementation is fine.

@ARUNMOHANRAJ471 can you grab a screenshot of a small plugin and a large plugin like I did above?

zbynek:
[Mailer](https://plugins.jenkins.io/mailer/)
image

[Last changes](https://plugins.jenkins.io/last-changes)
image

[Parallels Desktop](https://plugins.jenkins.io/parallels-desktop/)
image

halkeye:
It definitely highlight the changes a lot more but won't be great at showing the differences between plugins.

I'm honestly not sure what the goal of the graph is, so lets move forward with the new scale and see if anyone complains.

zbynek (approved):

src/components/LineChart.jsx
```diff
@@ -13,21 +13,13 @@ const calculateHeight = (total) => {
...
-    let maxValue = undefined;
-    let minValue = undefined;
-    if (lastValue < 100) maxValue = 250;
-    else if (lastValue < 250) maxValue = 500;
-    else if (lastValue < 500) maxValue = 1000;
-    else if (lastValue < 1000) maxValue = 2000;
-    else if (lastValue < 2500) maxValue = 5000;
-    else if (lastValue < 7500) maxValue = 10000;
-
-    if (lastValue > 5000) minValue = 0;
```
daniel-beck, line 26: Removing this is a bad idea, as the graphs are now very misleading. I've been so confused about the insane jumps in popularity in some plugins — how did https://plugins.jenkins.io/git-parameter/ **double** its installations!? Wait… it didn't. It's like +10% and graphs are now terrible.

If there is no easily visible difference in numbers if you don't show Y=0, then there are no changes that matter. The way we collect this data is unreliable in the first place, so trying to discern differences between a few percentages is not useful anyway.

daniel-beck (commented):
This should be reverted or overhauled. Not including Y=0 is actively detrimental when trying to understand the graphs.

src/components/LineChart.jsx
daniel-beck: Amended in #445.

Issue #137, opened 2020-02-27
Add a plugin tag cloud to the front page and search windows


oleg-nenashev:
It would be cool if we could use labels to generate a tag cloud. It would help to highlight the most popular areas/technologies and help users to discover them.

Nice 2 have: each label is a hyperlink

![image](https://user-images.githubusercontent.com/3000480/75437540-a47df880-5956-11ea-9b97-cb922fe02ed3.png)

tkelley-eg:
I spent just a little time on this and it seems this component would be nice:

https://github.com/madox2/react-tagcloud

This query should do the trick and can be reduced down to the data structure required by the component:

```javascript
query MyQuery {
  allJenkinsPlugin {
    nodes {
      labels
    }
  }
}
```
I hope to take a crack at the javascript code to translate the structure sometime this week.

daniel-beck:
Given what the actual tags that are used most often seem to be per the linked PR, I seriously doubt this would be all that useful. We're not highlighting all the cool technologies you can use with Jenkins, but mostly expose the exciting categories of "builder", "notifier", "external", "post-build", "report", and "misc".

Highlighting how popular the tags "adopt-this-plugin" and "deprecated" are is also not quite front page material.

oleg-nenashev:
We could filter out the tags we do not want there

tkelley-eg:
> Given what the actual tags that are used most often seem to be per the linked PR, I seriously doubt this would be all that useful. We're not highlighting all the cool technologies you can use with Jenkins, but mostly expose the exciting categories of "builder", "notifier", "external", "post-build", "report", and "misc".
> 
> Highlighting how popular the tags "adopt-this-plugin" and "deprecated" are is also not quite front page material.

It seems valuable that we've actually been able to see what the cloud looks like so there can be some debate on whether the information is useful or not.  I'm not sure what process you all might use to determine if this is a worthwhile feature for the front page or anywhere else in the app?  I'm happy to invest more time to complete this, but only if it's seems valuable enough to get merged :-)

It's not a problem to add an `block` list to prevent some less useful labels from appearing.  It would seem reasonable that the new react component could just store that locally.

cc: @halkeye @zbynek

halkeye:
> Highlighting how popular the tags "adopt-this-plugin" and "deprecated" are is also not quite front page material.

yea lets hide those tags. Once we get rid of the hard to read colors, i'm happy to merge it, and then revisit

tkelley-eg:
[...]

PR #404, opened 2020-10-18
Add Plugin Tag Cloud

tkelley-eg:
- Created new PluginLabelTagCloud component.
- Added graphql query to get labels and transform to structure required by new TagCloud component.
- Added component to index and need styling help.

Related to issue #137  

Summary of this pull request: 

I plopped this on the main page, but could use a little styling advice.  Is the vertical location good?  Should we create some margins for whitespace on the sides?

For the search page, should it be placed at the bottom on the left criteria area?

image


src/components/PluginLabelTagCloud.jsx
```diff
@@ -0,0 +1,50 @@
...
+            }
+            return obj;
+
+        }, {});
+
+        return prev;
+    }, {});
+
+    const tags = new Array();
+    Object.entries(allLabels).forEach(([value, count]) => {
```
tkelley-eg: It wasn't obvious to me how to create this structure while doing the `reduce` ops above....

zbynek: Maybe it can be simplified by converting the inner `reduce` to `forEach` (you don't really need `prevLabel` for anything)

halkeye: i think this is fine, but you could map it

```
const tags = Object.entries(allLabels).map(([value, count]) => { return { value, count } });
```

halkeye:
> I plopped this on the main page, but could use a little styling advice. Is the vertical location good? Should we create some margins for whitespace on the sides?

Do you have a screenshot for this? I'm so bad with whitespace but I'd like to see it in action (my dev environment is offline right now so i can't pull it)

> For the search page, should it be placed at the bottom on the left criteria area?

I'm not really sure it should go on the search page at all, just speaking out loud here, what about the main page, and a dedicated /ui/tagcloud page?

src/components/PluginLabelTagCloud.jsx
```diff
@@ -0,0 +1,50 @@
...
+        tags.push({value, count});
+    });
+
+    return (
+        <TagCloud
+            minSize={12}
+            maxSize={35}
+            tags={tags}
+            onClick={tag => document.location.href=`/ui/search/?labels=${tag.value}`}
```
halkeye: Instead of changing the location href, which will work, you should use the gatsby helper - https://www.gatsbyjs.com/docs/gatsby-link/#how-to-use-the-navigate-helper-function

This will prevent the entire page load if data is cached and other things.

src/components/PluginLabelTagCloud.jsx
```diff
@@ -0,0 +1,50 @@
...
+            labels: allJenkinsPlugin {
+                nodes {
+                  labels
+                }
+            }
+        }
+    `).labels.nodes.reduce((prev, nodes) => {
+
+        nodes.labels.reduce((prevLabel, obj) => {
+            if (!prev[obj]) {
```
halkeye: just fyi, i'm not sure its any clearer to read, you can do it in one line

```
prev[obj] = (prev[obj] || 0) + 1
```

zbynek (commented):
Thanks for the PR. Maybe @oleg-nenashev can comment on the desired position of this tag cloud. Depending on the position we should adjust the colors used by the cloud -- currently it's too light to be readable on white background.

zbynek:
@halkeye  there's the screenshot:

image

For margins it would be better to apply some existing classname from Bootstrap; I think `container` should do the trick.

halkeye:
> @halkeye there's the screenshot:

oh thats slick. Some of the colors are hard on the eyes on the white bg, and I think it needs context, but its a nice addition

tkelley-eg:
Is there anything I might do to help bring this one across the finish line?  Thanks!

zbynek:
Given @daniel-beck 's comment I'm not sure what's the best way forward here -- there are many more labels that should be filtered (all labels that are used for a single plugin, labels like `api-plugin`, probably also `builwrapper`, `post-build`, `ui`, `slaves`, `agent`...), in some cases synonyms should be merged (stash = bitbucket = bitbucket-server). Maybe what we need to achieve @oleg-nenashev 's original goal (highlighting useful plugins from different areas) is not a word cloud at all.

I think it's good that thanks to this PR we have an idea about the used labels and their frequencies, but maybe the UX sig could chime in before you spend time perfecting a feature that's gonna be removed.

oleg-nenashev:
Sorry. It was not in my review list and I have missed a few pings. Will try to review it over the weekend and come up with suggestions. 

I envisioned a relatively small tag cloud pane under the existing quick filters, something like that:

image
.

Maybe we could also put it under the search to utilize the currently unused space on desktop resolutions

image

src/components/PluginLabelTagCloud.jsx
```diff
@@ -0,0 +1,51 @@
...
+                prev[obj] = (prev[obj] || 0) + 1;
+            }
+            return obj;
+
+        }, {});
+
+        return prev;
+    }, {});
+
+    const tags = Object.entries(allLabels).map(([value, count]) => {return {value, count};});
```
fqueiruga, line 31: `Object.entries` is not supported on IE https://caniuse.com/object-entries.  I think a polyfill is needed or use `for .. in` instead..

fqueiruga (changes requested):

src/components/PluginLabelTagCloud.jsx
halkeye: everything in gastby will go through babel so shouldn't need to worry.

That being said, one of the dependancies breaks on ie10 because it uses Map and doesn't seem to be babeled - https://app.saucelabs.com/tests/b9d8dc52f6354636b481ca0f73f85a22
Looks like also breaks in ie11, but different place, somewhere in react - https://app.saucelabs.com/tests/64266ab032ab42f181a9b1131e7db982

That being said, the site is designed to export regular html with javascript applied on top after its finished loading, so the older browsers won't get the full experience, but they will get the info. I wouldn't worry about object.entries

halkeye:
For those who want to see it live:
Unique Deploy URL: https://5f9c4ccad4709a0c81d4fded--pensive-galileo-3fa514.netlify.app

zbynek:
This has been inactive for a long time, if someone strogly feels this is worth implementing, feel free to reopen.

PR #445, opened 2020-10-30
Make popularity graph start at y=0

daniel-beck:
Amends #402 to always start the graph at y=0. Showing a narrower range is misleading at a glance. Even if it shows small differences more clearly, the downside doesn't seem worth it.

FYI @ARUNMOHANRAJ471 

# Example: Showing rapidly changing popularity

While both https://plugins.jenkins.io/trilead-api/ and https://plugins.jenkins.io/okhttp-api/ gained popularity, the relative gains are very different, so should result in different graphs.

## Before

trilead-api
okhttp-api

## After

trilead-api
okhttp-api

# Example: Showing fairly unchanging popularity

Additionally, this will now not over-emphasize fairly insignificant differences in reported installation counts.

## Before

matrix-auth
git-parameter

## After

matrix-auth
git-parameter

### DIFF
--- a/src/components/LineChart.jsx
+++ b/src/components/LineChart.jsx
@@ -13,12 +13,11 @@ const calculateHeight = (total) => {
 };
 
 const calculateMinMax = (data) => {
-    const minValue = Math.min(...data);
     const maxValue = Math.max(...data);
     // calculate a dynamic value to center the graph
     const scaleDifference = Math.pow(10, maxValue.toString().length-1);
     return {
-        min: parseInt(minValue/scaleDifference)*scaleDifference,
+        min: 0,
         max: parseInt((maxValue/scaleDifference)+1)*scaleDifference // plus 1 to add more space in the top
     };
 };


src/components/LineChart.jsx
```diff
@@ -13,12 +13,11 @@ const calculateHeight = (total) => {
 };
 
 const calculateMinMax = (data) => {
-    const minValue = Math.min(...data);
     const maxValue = Math.max(...data);
     // calculate a dynamic value to center the graph
```
daniel-beck, line 17: We're no longer _centering_ things here, so I'm not sure how well this comment applies anymore.

daniel-beck:
The "worst case" for the range algorithm is when the max install count is just over a power of 10. Then the top of the graph is 2x that power of 10, like for `TestComplete`:

after

The current behavior isn't great either though (IMO), I would even argue there's an improvement:

before

zbynek:
This is what I was suggesting in the original PR but I thought I was in the minority on that. Thanks for the PR!

daniel-beck:
@zbynek Thanks, I appreciate the agreement and merge – although it might be better to give potentially controversial PRs like this a little more time to collect feedback.

PR #484, opened 2020-11-22
Upgrade Releases tab rewith the update center info then augment with release notes

halkeye:
Adding a archives tab which lists past releases.

Its a fairly early pass and UI is not my specialty


halkeye:
~Demo: https://jenkins-plugins.g4v.dev~

halkeye:
Demo: https://plugin-site-regud.ondigitalocean.app/

src/components/PluginReleases.jsx
```diff
@@ -2,52 +2,65 @@ import React, {useState, useEffect} from 'react';
...
+                                    <TimeAgo date={new Date(version.buildDate)} formatter={formatter}/>
                                 </div>
                             </h5>
                         </div>
                         <div className="card-body">
-                            <p
+                            <div>
+                                <InstallViaCLI pluginId={pluginId} version={version.version} />
+                            </div>
+                            {release.bodyHTML && (<blockquote
```
timja, line 56: I'm not a huge fan of this block quote bit, I think we should remove it,

especially I don't like the italics
https://plugin-site-regud.ondigitalocean.app/slack/#releases

image

halkeye: > I'm not a huge fan of this block quote bit, I think we should remove it,

and replace it with? just the \<p\> directly doesn't really work, it blends everything together, though open to ideas with spacing and stuff

timja: what if we just move the install via cli to after the notes and before the checksums?

halkeye:
Merging, and we'll revisit the design in a future PR

src/templates/plugin.jsx
```diff
@@ -73,14 +73,13 @@ function PluginPage({data: {jenkinsPlugin: plugin, reverseDependencies: reverseD
...
                             {plugin.wiki.content && <div className="content" dangerouslySetInnerHTML={{__html: plugin.wiki.content}} />}
                         </>)}
-                        {state.selectedTab === 'releases' && <PluginReleases pluginId={plugin.id} />}
+                        {state.selectedTab === 'releases' && <PluginReleases pluginId={plugin.id} versions={versions.edges.map(edge => edge.node)} />}
                         {state.selectedTab === 'issues' && <PluginIssues pluginId={plugin.id} />}
                         {state.selectedTab === 'dependencies' && <PluginDependencies dependencies={plugin.dependencies} reverseDependencies={reverseDependencies.edges}/>}
                     </div>
                 </div>
                 <div className="col-md-3 gutter">
-                    <a href={`https://updates.jenkins.io/download/plugins/${plugin.name}`}
```
daniel-beck, line 82: A second redundant link to the same tab is redundant.

Finding this out during advisory publication is 👎

PR #497, opened 2020-11-30
Group reverse dependencies by type

zbynek:
The reverse dependencies are currently listed without indication which one is reqired/optional/implied.
Also reverse dependencies are not shown at all if plugin has no dependencies. Both issues are fixed here.

I didn't manage to find a GraphQL query that would work using the JenkinsPlugin type (and would not slow down everything by grabbing too much data), so I aded JenkinsPluginDependency. It's not very DRY but since the data is just used internally it should not matter.


markjacksonfishing (approved):

plugins/gatsby-source-jenkinsplugins/utils.js
```diff
@@ -115,6 +115,20 @@ const fetchPluginData = async ({createNode, reporter}) => {
                         contentDigest: crypto.createHash('md5').update(`plugin${pluginData.name.trim()}`).digest('hex')
                     }
                 });
+                pluginData.dependencies.forEach(dependency => {
+                    const mergedId = `${pluginData.name.trim()}:${dependency.name.trim()}`;
+                    createNode({
```
halkeye: so createNode returns a promise (https://www.gatsbyjs.com/docs/actions/#createNode) that means all these dependency nodes won't be waiting until completion.


```suggestions
                createNode().then(pluginNode => Promise.all(pluginData.dependencies.map(dependency => {
                    const mergedId = `${pluginData.name.trim()}:${dependency.name.trim()}`;
                    return createNode({});
                  });
```

halkeye:
Btw, if you want to get an auto buildable for your PR like I have for #484 via digital ocean apps (free for static sites)

UI:
image

Yaml:
[plugin-site.txt](https://github.com/jenkins-infra/plugin-site/files/5620497/plugin-site.txt)

halkeye (approved):

halkeye:
@zbynek CI is failing:

image

halkeye:
oh createnode mock isn't returning a promise :D

PR #535, opened 2020-12-19
Make sure PR builds are not rebuilt periodically

zbynek:
Jenkinsfile_k8s is used for deployment, Jenkinsfile just for CI, so it doesn't need the deploy stage and more importantly the cron trigger (no need to rebuild PRs).

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -8,10 +8,6 @@ pipeline {
     label 'docker&&linux'
   }
 
-  triggers {
-    cron('H H/3 * * *')
-  }
-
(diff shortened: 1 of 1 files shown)


markjacksonfishing (approved):

timja (approved):

halkeye:
I was actually thinking I'd disable it except major and security updates. But let's try this first cause it's better

halkeye:
I suspect we'll have to do trigger {} to actually force the cron to stop, but merging for now and we'll see

PR #580, opened 2021-02-21
Switch to using optional Algolia Search

halkeye:
Demo: https://plugin-site-regud.ondigitalocean.app/ui/search?query=blue%20ocean
Discussion: https://groups.google.com/g/jenkinsci-docs/c/B8vREtrD0Uw


MarkEWaite (approved):
I am not experienced with JavaScript or Algolia, but I experimented with the test site and it was well behaved in all the things that I checked.  I checked:

* bitbucket (new is better because it does not include GitHub results like current does)
* markewaite (better on new site, knows me)
* git (comparable)
* gitea (more results from new site, but they were GitLab results rather than gitea)
* github (better results from the new site)

plugins/gatsby-source-jenkinsplugins/utils.js
```diff
@@ -115,7 +115,23 @@ const fetchPluginData = async ({createNode, reporter}) => {
                         type: 'JenkinsPlugin',
                         contentDigest: crypto.createHash('md5').update(`plugin${pluginData.name.trim()}`).digest('hex')
                     }
-                }).then(() => {
+                });
+                if (!p || !p.then) {
```
zbynek, line 119: Can `createNode` return something else than a valid promise?

halkeye: It shouldn't. I was going to look into it, but it turned a random null once in a while for me

plugins/gatsby-source-jenkinsplugins/utils.js
```diff
@@ -115,7 +115,23 @@ const fetchPluginData = async ({createNode, reporter}) => {
...
+                    console.log('p', p, plugin);
```
timja, line 127: is this supposed to still be here? debug?

timja (approved):
looks great

plugins/gatsby-source-jenkinsplugins/utils.js
halkeye: Hmmmm.. in theory it should always return a promise but sometimes it doesn't. So wanted to track it down.

Thinking about it now maybe I should sentry it. Next pr will be doing a mass upgrade of all deps so hopefully it'll fix it but I guess sentry is easier to debug than VPN locked ci

markjacksonfishing (approved):

zbynek:
image
The alignment of the search logo is strange, on certain screen sizes part of the logo is cut off. Maybe it could be moved to the next  row where the active filters are (or if license permits, below the search results and centered).

zbynek:
The number of installs is missing on the cards in search.

halkeye:
> The alignment of the search logo is strange, on certain screen sizes part of the logo is cut off.

Drat i knew it was too easy. I only tested small and desktop sizes. 

> or if license permits, below the search results and centered

"The Free Plan requires you to display the [Algolia logo](https://www.algolia.com/press/?section=brand-guidelines) next to the search results."
"In exchange for using Algolia for free, we ask that you place a small "Search by Algolia" logo in your search results or on your site where people can see it. The logo should link to https://algolia.com. If this is not possible or ideal, there are other ways to fulfill this requirement that we're happy to discuss with you."

I'll try out the new row, or smaller image when I get a minute to update this

halkeye:
> The number of installs is missing on the cards in search.

fixed, had code from before i copied the entire stats object over

plugins/gatsby-source-jenkinsplugins/utils.js
halkeye: Its going to sentry now

halkeye:
> 
> 
> image
> The alignment of the search logo is strange, on certain screen sizes part of the logo is cut off. Maybe it could be moved to the next row where the active filters are (or if license permits, below the search results and centered).

I think below isn't in the spirit of the request.

image

timja (approved):

PR #588, opened 2021-02-25
Switch to using react-paginate so we dont need to manage that code ourselves 

halkeye:
Related to issue # 586

Summary of this pull request:


markjacksonfishing (approved):

MarkEWaite (approved):
Really like the switch to react paginate so that someone else can do the computation of page components.

PR #618, opened 2021-04-01
Add synonyms for p4 and perforce

halkeye:
As reported on irc:

> @omehegan: @halkeye: I just noticed that when you search the plugins page for 'perforce' all the results are as if I searched for 'performance.' neither of the perforce plugins show up :) https://plugins.jenkins.io/ui/search?query=perforce
> @halkeye: autojack: https://github.com/jenkins-infra/update-center2/blob/1690b80d6a92489c1be9ba51105c47ba4f3e5d9d/resources/artifact-ignores.properties#L319 short version is perforce is blacklisted
> @omehegan: got it. but if someone wants to find the p4 plugin, they have to know to search for p4?
it just seems like searching for perforce should at least return the p4 plugin.
> @halkeye: oh interesting, we ended up disabling searching the body of the readme file
> @halkeye: okay i can add it as a synonym, it'll find perforc and perforce, but not perfor which feels right
> @omehegan: sounds good! just one of those weird gotchas.
> @omehegan: I was looking at the results, very confused. "This is almost, but not quite, exactly what I wasn't looking for."

I like everything config inside the code, but that does mean that anything we do in the UI will get overridden, so we'd have to decide now if we want synonyms in there


timja (approved):

MarkEWaite (approved):
Yes, that's a great idea.

PR #646, opened 2021-05-01
Fix typo in dependabot config

zbynek:
See https://dependabot.com/docs/config-file/#ignored_updates


.github/dependabot.yml
```diff
@@ -5,6 +5,6 @@ updates:
   schedule:
     interval: monthly
   open-pull-requests-limit: 10
-  ignore_updates:
+  ignored_updates:
```
MarkEWaite: I don't see any description for either `ignore_updates` or `ignored_updates` in the [dependabot docs](https://docs.github.com/en/code-security/supply-chain-security/configuration-options-for-dependency-updates).  The Jenkins core [dependabot config file](https://github.com/jenkinsci/jenkins/blob/53823453a276de2e4fff20ec0b15c91aa806a0e3/.github/dependabot.yml#L11) uses `ignore` as do the plugins where I've seen it used.

Issue #629, opened 2021-04-07
Deprecated Plugins can give 403 errors


halkeye:
For example https://plugins.jenkins.io/azure-iot-edge/ gives an nginx 403 forbidden error. Our unverified guess is that the directory exists, but the file got deleted when the plugin was removed. 

Easiest solution, process https://github.com/jenkins-infra/update-center2/blob/master/resources/deprecations.properties and create a depreciated page and use the link in the properties file if available.

Recommendation from writer:
* Create new JenkinsPluginDeprecations in https://github.com/jenkins-infra/plugin-site/blob/master/plugins/gatsby-source-jenkinsplugins/utils.js
* Create new page similar to https://github.com/jenkins-infra/plugin-site/blob/adfee9e12b34bf0bdbfa0923d6d21b6c4138297d/gatsby-node.js#L56-L61

PR #669, opened 2021-05-08
Add tombstone page for suspended plugins

zbynek:
Related to issue #629

Summary of this pull request: simple tombstone page for plugins that were removed from distribution.
image

May be extended later by providing human-readable title of the plugin and maybe some other metadata (a link to GitHub repo in case someone wants to revive the plugin?), but that would require changes in other components.

CC @daniel-beck


plugins/gatsby-source-jenkinsplugins/utils.js
```diff
@@ -159,6 +161,26 @@ const fetchPluginData = async ({createNode, reporter}) => {
         page = pluginsContainer.page + 1;
     } while (!page || pluginsContainer.page < pluginsContainer.pages);
     await Promise.all(promises);
+    const updateUrl = 'https://updates.jenkins.io/current/update-center.actual.json';
+    const updateData = await requestGET({url: updateUrl, reporter});
+    const suspendedPromises = [];
+    for (const deprecation of Object.keys(updateData.deprecations)) {
+        if (names.indexOf(deprecation) == -1) {
```
halkeye: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes
I recommend `names.includes(deprecation)` instead of indexof. Also I think performance wise, a set is faster than a straight up array, but i doubt it matters for this.

plugins/gatsby-source-jenkinsplugins/utils.js
```diff
@@ -159,6 +161,26 @@ const fetchPluginData = async ({createNode, reporter}) => {
...
+            suspendedPromises.push(createNode({
+                wiki: updateData.deprecations[deprecation],
+                id: deprecation,
+                name: deprecation,
+                suspended: true,
+                parent: null,
+                children: [],
+                internal: {
+                    type: 'JenkinsPlugin',
```
halkeye: Lets make this a completely different type. It only needs name, id and url right? then you can just loop through the new type in gatsby-node.js

Just simplifies things and keeps them seperate

plugins/gatsby-source-jenkinsplugins/utils.js
```diff
@@ -159,6 +161,26 @@ const fetchPluginData = async ({createNode, reporter}) => {
         page = pluginsContainer.page + 1;
     } while (!page || pluginsContainer.page < pluginsContainer.pages);
     await Promise.all(promises);
+    const updateUrl = 'https://updates.jenkins.io/current/update-center.actual.json';
```
daniel-beck:
```suggestion
    const updateUrl = 'https://updates.jenkins.io/update-center.actual.json';
```

Better to not rely on this implementation detail.

plugins/gatsby-source-jenkinsplugins/utils.js
```diff
@@ -159,6 +161,26 @@ const fetchPluginData = async ({createNode, reporter}) => {
...
+    const updateData = await requestGET({url: updateUrl, reporter});
+    const suspendedPromises = [];
+    for (const deprecation of Object.keys(updateData.deprecations)) {
```
daniel-beck, line 167: This does not actually list plugins whose distribution is suspended, but those marked deprecated -- this feature exists just so that plugins can be marked deprecated while being suspended, but having the customizable URL is a benefit that could result in plugins being marked deprecated through this feature even while still being distributed.

Some plugins have been removed without having an entry here (e.g. `emotional-hudson`). Other plugins may have an entry here without being suspended (unlikely, but as I wrote, possible as it has a benefit).

At a minimum, this needs to cross-check that no plugin found here is still actively distributed before just showing the tombstone page. Then we can just accept that _some_ (few) plugins still don't show up.

The alternative is parsing https://github.com/jenkins-infra/update-center2/blob/master/resources/artifact-ignores.properties and relying on ugly internals of that tool, or extending it to list what's suspended in an additional output file. As a start, the current approach is pretty reasonable if we ensure no currently distributed plugin would have a tombstone page.

daniel-beck (commented):
Thanks for taking care of this!

oleg-nenashev:
It sis definitely a great start! Just having a tombstone will be very helpful so that plugin links are preserved. We can iterate on that in subsequent pull requests

plugins/gatsby-source-jenkinsplugins/utils.js
zbynek: > Other plugins may have an entry here without being suspended

there is a check for the false positives already. For false negatives we can manually check the `artifact-ignores.properties` and add some more entries to the deprecations if needed (if it's not worth showing as deprecated in Jenkins, it probably doesn't need a tombstone page either).

halkeye (approved):
Looks good to me

plugins/gatsby-source-jenkinsplugins/utils.js
daniel-beck: Great, then this is good to go from my POV :+1:

src/templates/tombstone.jsx
```diff
@@ -0,0 +1,54 @@
...
+                        {plugin.id}
+                    </h1>
+                    <div className="plugin-id">
+                        {'ID: '}
+                        {plugin.id}
+                    </div>
+                </div>
+                <div className="alert alert-warning alert-with-icon">
+                    <WarningsIcon/>
+                    {'Plugin was removed from distribution, see '}
```
daniel-beck: This would align with the wording I generally use when talking about this. Seems less final (which in many cases it is not).

```suggestion
                    {'Plugin distribution has been suspended, see '}
```

daniel-beck (approved):

plugins/gatsby-source-jenkinsplugins/utils.js
zbynek: plugins with more than 100 installs (as of April 1) that don't have a deprecation entry:
```
7205 copy-to-slave
5930 build-flow-plugin
4691 tfs
3093 dynamicparameter
1629 perforce
1276 svn-tag
1047 kubernetes-pipeline-steps
761 scripttrigger
595 grails
559 kubernetes-pipeline-aggregator
558 kubernetes-pipeline-arquillian-steps
435 kubernetes-ci
367 zap-pipeline
333 persona
278 jshint-checkstyle
230 xtrigger
229 play-autotest-plugin
200 pipeline-classpath
188 puppet-enterprise-pipeline
184 simple-travis-runner
181 groovyaxis
162 gcm-notification
152 cvs-tag
148 fortify360
101 jslint-checkstyle
```

PR #675, opened 2021-05-13
Simplify CSS (remove main.module.css)

### LINKED ISSUE #80: Completely remove the main.module.css
main.module.css is using css modules, and a lot of !important rules

Migrate everything to normal postcss style css rules so its using the same system

zbynek:
Fixes #80  (see issue for details)


timja (approved):

halkeye:
hrmmmmm, at some point when I created #80, I was really annoyed at the !importants, and eventually I came around on the css modules.

I'm absolutely cool with this PR once its updated with latest master, but we should also decide if we want to do modules or not. There's 3 other files that I could pull out in a followup PR.

https://github.com/jenkins-infra/plugin-site/blob/master/src/components/Checksum.module.css
https://github.com/jenkins-infra/plugin-site/blob/master/src/components/InstallViaCLI.module.css
https://github.com/jenkins-infra/plugin-site/blob/master/src/components/SearchByAlgolia.module.css

The nice thing about css modules is that they are contained and a module doesn't care where its used, it just has css applied. But at the same time, it makes it harder to see whats going on.

So feelings?

halkeye (approved):

src/components/Categories.jsx
```diff
@@ -26,7 +24,7 @@ function Categories({clearCriteria, toggleCategory, activeCategories, activeLabe
     const anyCriteria = activeCategories.length > 0 || activeLabels.length > 0;
     
     return (
-        <fieldset className={classNames(styles.Categories)}>
+        <fieldset className={'Categories'}>
```
halkeye:
```suggestion
        <fieldset className="Categories">
```

src/components/Categories.jsx
```diff
@@ -35,7 +33,7 @@ function Categories({clearCriteria, toggleCategory, activeCategories, activeLabe
                     </button>
                 )}
             </legend>
-            <ul className={classNames(styles.Cols3, 'Cols3')}>
+            <ul className={'CategoryList'}>
```
halkeye:
```suggestion
            <ul className="CategoryList">
```

halkeye:
a couple minor nitpicky stuff, not blocking the {''} to "" stuff, i just feel its cleaner

zbynek:
I think modules are fine if they have clearly defined responsibilities, i.e. classes defined in the module are not present in base.css and the module dos not contain classes for unrelated components. Maybe we could do modules for new components and keep the non-module CSS files for old stuff. Also we're loading bootstrap by default, so we can leverage some of the classes it defines like `mt-3` and avoid creating own classes just for setting margin.

PR #754, opened 2021-07-05
Switch labels, categories and detachments to UC

zbynek:
Reduce the dependency on `plugin-site-api`
* move categories here as a static file (currently a static file in `plugin-site-api`)
* get detachments directly from Jenkins source via GitHub (currently using GitHub also in `plugin-site-api`)
* get labels directly from Jenkins source via GitHub (currently a static file in `plugin-site-api`, uses slightly differerent wording)

### DIFF
--- a/plugins/gatsby-source-jenkinsplugins/utils.js
+++ b/plugins/gatsby-source-jenkinsplugins/utils.js
@@ -96,8 +96,6 @@ const getPluginContent = async ({plugin, reporter}) => {
     });
 };
 
-const unique = (value, index, self) => self.indexOf(value) === index;
-
 const fetchPluginData = async ({createNode, reporter, firstReleases, labelToCategory}) => {
     const sectionActivity = reporter.activityTimer('fetch plugins info');
     sectionActivity.start();
@@ -130,7 +128,7 @@ const fetchPluginData = async ({createNode, reporter, firstReleases, labelToCate
                     securityWarnings: updateData.warnings.filter(p => p.name == pluginName)
                         .map(w => checkActive(w, pluginUC)),
                     dependencies: pluginUC.dependencies.concat(getImpliedDependencies(plugin, detachedPlugins, updateData)),
-                    categories: pluginUC.labels.map(label => labelToCategory[label]).filter(unique),
+                    categories: Array.from(new Set(pluginUC.labels.map(label => labelToCategory[label]))),
                     firstRelease: firstReleases[pluginName].toISOString(),
                     id: pluginName,
                     hasPipelineSteps: pipelinePluginIds.includes(pluginName),
--- a/plugins/gatsby-source-jenkinsplugins/utils.js
+++ b/plugins/gatsby-source-jenkinsplugins/utils.js
@@ -96,7 +96,7 @@ const getPluginContent = async ({plugin, reporter}) => {
     });
 };
 
-const fetchPluginData = async ({createNode, reporter, firstReleases, labelToCategory}) => {
+const fetchPluginData = async ({createNode, reporter, firstReleases}) => {
     const sectionActivity = reporter.activityTimer('fetch plugins info');
     sectionActivity.start();
     const promises = [];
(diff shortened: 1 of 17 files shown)


plugins/gatsby-source-jenkinsplugins/utils.js
```diff
@@ -7,6 +7,7 @@ const URL = require('url');
 const axiosRetry = require('axios-retry');
 const dateFNs = require('date-fns');
 const {parseStringPromise} = require('xml2js');
+const categoryList = require('./categories.json');
```
halkeye, line 10: I'm not totally sold on plugin site being the owner of any data... I'll be honest I also don't think categories are very useful right now. How do you feel about getting the static json file inside of update center?

plugins/gatsby-source-jenkinsplugins/utils.js
```diff
@@ -122,8 +129,8 @@ const fetchPluginData = async ({createNode, reporter, firstReleases}) => {
                     wiki: pluginData.wiki,
                     securityWarnings: updateData.warnings.filter(p => p.name == pluginName)
                         .map(w => checkActive(w, pluginUC)),
-                    dependencies: pluginData.dependencies,
-                    categories: pluginData.categories,
+                    dependencies: pluginUC.dependencies.concat(getImpliedDependencies(plugin, detachedPlugins, updateData)),
+                    categories: pluginUC.labels.map(label => labelToCategory[label]).filter(unique),
```
halkeye: Alternatively:
Array.from(new Set(pluginUC.labels.map(label => labelToCategory[label])))

halkeye: if we are just mapping labels to categories, we don't really need categories in plugins at all do we? Couldn't it just be a graphql query on the displaying plugin page?

plugins/gatsby-source-jenkinsplugins/utils.js
zbynek: I could move it there of course, but I suspect it's out of scope for update center too? @daniel-beck ideas?

plugins/gatsby-source-jenkinsplugins/utils.js
zbynek: kindof, it would also need some refactoring for the Algolia index I suppose

zbynek: actually the search index is the only place where we need a connection between plugins and categories.

halkeye: oh interesting, so we could in theory just say hey algoila, we want anything with labels x y and z and not worry about linking it to a specific plugin? thats cool

zbynek: done in the last commit. So, if the whole category thing is just to define grouping of labels in the search form, maybe it's OK to keep the category definition in this repo?

halkeye:
Yea. Another thing to add to the metric collecting to see if people use it.
But yea if it's only for search and not display anywhere I don't feel the
same anxiety that people will use the data. I don't like the idea of plugin
site being authority of anything.

halkeye:
I'll be on my desktop tonight and do the final review. I try not to be
logged into GitHub on my phone

src/templates/search.jsx
```diff
@@ -1,8 +1,9 @@
+
 // import {graphql} from 'gatsby';
 import React from 'react';
 import querystring from 'querystring';
 import PropTypes from 'prop-types';
-import {navigate} from 'gatsby';
+import {navigate,useStaticQuery, graphql} from 'gatsby';
```
halkeye: minor nit about consistent spacing. I'm actually surprised there's no linting rule enabled for it

can you change it to either
```suggestion
import {navigate,useStaticQuery,graphql} from 'gatsby';
```
or
```suggestion
import {navigate, useStaticQuery, graphql} from 'gatsby';
```

src/templates/search.jsx
```diff
@@ -17,16 +18,19 @@ import Filters from '../components/Filters';
...
+const doSearch = (data, setResults, categoryList) => {
     const {query, sort} = data;
     let {categories, labels, page} = data;
     if (!Array.isArray(categories)) { categories = [categories]; }
     categories = categories.filter(Boolean);
     if (!Array.isArray(labels)) { labels = [labels]; }
     labels = labels.filter(Boolean);
-
+    for (const categoryId of categories) {
+        categoryList.edges.filter(e => e.node.id == categoryId).forEach(e => labels.push(...e.node.labels));
```
halkeye, line 32: oh awesome, this will make it so easy to add tracking later.

halkeye (approved):
lgtm

PR #759, opened 2021-07-10
Fix GitHub URLs for submodules

### LINKED ISSUE #520: Bitbucket Pipeline for Blue Ocean page - Invalid link
Problem with the [Bitbucket Pipeline for Blue Ocean](https://plugins.jenkins.io/blueocean-bitbucket-pipeline) page, [source file](https://github.com/jenkins-infra/plugin-site/tree/master/src/templates/plugin.jsx)

The reference to the corresponding Github project is invalid.

## Screenshots 

image

## Possible Solution 

Add a valid Github link, e.g. https://github.com/jenkinsci/blueocean-plugin/

zbynek:
Fixes #520

If GitHub URL is in the form github.com/org/repo/path, we should point to github.com/org/repo/tree/defaultBranch/path . Probably makes sense to normalize here instead of fixing in plugins because it's unclear what URLs should be in the POM of a plugin. Also it's not just the ~20 BO plugins, but also ~10 other plugins including https://plugins.jenkins.io/structs/

### DIFF
--- a/plugins/gatsby-source-jenkinsplugins/utils.js
+++ b/plugins/gatsby-source-jenkinsplugins/utils.js
@@ -123,6 +123,7 @@ const fetchPluginData = async ({createNode, reporter, firstReleases}) => {
             names.push(pluginName);
             const pluginUC = updateData.plugins[pluginName];
             pluginUC.developers.forEach(maint => {maint.id = maint.developerId, delete(maint.developerId);});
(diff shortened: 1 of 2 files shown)


halkeye:
~I think this should go into plugin site api not plugin site so it picks up readmes...~
readmes are a different field and thus not applicable

so LGTM

halkeye (approved):

PR #762, opened 2021-07-11
add @zbynek  to code owners

halkeye:
As a non org member i can't add @zbynek to the plugin site team, so hoping this will work in the mean time


oleg-nenashev (approved):
Is there an INFRA ticket for adding @zbynek ? I have no access as well, but a ticket is needed so that @olblak sees the request in the queue

olblak:
I sent an invitation to @zbynek to join the pluginsite team which I prefer

PR #764, opened 2021-07-28
Get stats from stats.jenkins.io

zbynek:
Further reduces usage of `plugin-site-api` by getting stats from stats.jenkins.io and documentation URLs from update center. The only remaining usage of `plugin-site-api` is for getting the GitHub docs (Confluence docs are handled directly since some time).


plugins/gatsby-source-jenkinsplugins/utils.js
```diff
@@ -115,76 +108,76 @@ const fetchPluginData = async ({createNode, reporter, firstReleases}) => {
...
+            });
+            if (!p || !p.then) {
+                if (process.env.GATSBY_SENTRY_DSN) {
+                    const Sentry = require('@sentry/node');
+                    Sentry.init({
+                        dsn: process.env.GATSBY_SENTRY_DSN
+                    });
+                    Sentry.captureMessage(new Error(`Error creatingNode for plugin: ${pluginName}`), {extra: {pluginData: plugin, p: p}});
+                }
+                //console.log('p', p, plugin);
```
timja:
```suggestion
```

timja (approved):

src/components/PluginReadableInstalls.jsx
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 
 function PluginReadableInstalls({currentInstalls}) {
-    if (!currentInstalls) {
+    if (!currentInstalls && currentInstalls !== 0) {
```
MarkEWaite, line 5: Does `!==` work as a comparison operator?  https://jsx.github.io/doc/operatorref.html does not mention it as a comparison operator
```suggestion
    if (!currentInstalls && currentInstalls != 0) {
```

halkeye: You should always use the triple version as it checks types and doesn't do boolean conversions (so false will never === 0 but will == 0)

halkeye:
so with the information we learned yesterday i'm reluctant to sign off on this (havn't finished a review anyways)

https://groups.google.com/g/jenkinsci-dev/c/r5kE8YEbSOE

A while ago i looked into fully removing plugin site api (unlike the kickass way your doing it one at a time) so I did do pulling the stats from git like plugin-site-api does, it might be a better route - https://github.com/halkeye/plugin-site/blob/migrate-from-plugin-site-api/plugins/gatsby-source-jenkinsplugins/utils.js#L100-L151

zbynek:
@halkeye I wanted to avoid git pull which takes 3minutes if I can do it wiith 13 HTTP requests that are ~1s total. Also not having to deal with Git dependency is nice. 

I assume the GH functionality syncing `gh-pages` branch with stats.jenkins.io is quite reliable, so it makes no difference if we fetch a file over HTTP or git pull. It might make a difference whether we take the CSVs or the JSONs from that branch, but so far I found no discrepancy between those.

The only problem seems to be the HTML files. They are also generated monthly, but
* the JEP-229 versions are included in the source of those files (see https://stats.jenkins.io/pluginversions/adobe-cloud-manager.html ), but are not shown in the table because of
https://github.com/jenkins-infra/infra-statistics/blob/gh-pages/pluginversions/pluginversions.js#L12
* there is a small discrepancy between the CSV (32 installs total) and the data inside the HTML (31 total), because the HTML is excluding `1.0.3-SNAPSHOT (private)`

EDIT: problem with JEP-229 plugin versions hopefully fixed here https://github.com/jenkins-infra/infra-statistics/pull/47

zbynek:
@halkeye with https://github.com/jenkins-infra/infra-statistics/pull/47 resolved, are you still concerned about the approach here?

halkeye:
Not concerned just haven't reviewed. If your confident merge

PR #790, opened 2021-08-13
Use deprecation notice URLs where available 

### LINKED ISSUE #763: Add support for custom deprecation URLs
Followup to #752: The deprecations map provides a custom URL that is expected to explain in some manner why a plugin is deprecated. This is generally useful if plugins are no longer maintained or distributed (see also https://github.com/jenkins-infra/update-center2/#deprecations ).

In Jenkins, the admin monitor will link there, or (if deprecated using label) to the plugin URL (which is just the plugin site at the moment).

Doing the latter is pointless on the plugin site of course, but if a deprecation is done through the deprecations map, it would be useful to add link there in the warning.

As an example, https://plugins.jenkins.io/harvest/ is such a plugin and deprecation, the message could link to the core PR that makes Jenkins incompatible with the plugin: https://github.com/jenkins-infra/update-center2/blob/f4d86c1a6278ffd5bb4efc1f85f439b6cf3e46f4/resources/deprecations.properties#L103-L104

zbynek:
Fixes #763 

Plugin with no URL
image

Plugin with URL
image

CC @daniel-beck

### DIFF
--- a/plugins/gatsby-source-jenkinsplugins/utils.js
+++ b/plugins/gatsby-source-jenkinsplugins/utils.js
@@ -113,7 +113,11 @@ const fetchPluginData = async ({createNode, reporter, firstReleases}) => {
     const detachedPlugins = detachedPluginsData.split('\n')
         .filter(row => row.length && !row.startsWith('#')).map(row => row.split(' '));
     for (const deprecation of Object.keys(updateData.deprecations)) {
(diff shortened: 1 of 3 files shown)


halkeye (approved):
lgtm

daniel-beck (approved):
Thanks a lot!

PR #791, opened 2021-08-31
Search UI improvements

### LINKED ISSUE #82: Fix pagination when changing filters
Going to page 2 of search results, then searching for more specific (adding labels/categorys/or query string) will show no results cause on page 2 not page 1

zbynek:
* Fixes #82 
* Makes query handling consistent  (currently buggy, e.g. search by query, add filter, remove query, remove filter => results still filtered by query)
* CSS fix for list view (make sure it's distinguishable from cards view)
* Show release date instead of required core in results

Before
search-before

After
search-after

("released 3 minutes ago" in the screenshot is just because Algolia data needs to be updated, should be also covered by the PR)


halkeye:
sorry with my increased hosting duties i'm letting things slip more. Its a
long weekend for me so hopefully I can catch up.'

halkeye:
Seems fine. Lets try it.

PR #808, opened 2021-09-15
Use plugin name instead of ID where needed

zbynek:
In the recent refactoring `id` became Gatsby internal thing and `name` is the correct way to get the plugin id/name/slug or how should we call it.

Because of this the release tab is not showing any release details, the CLI install strings are wrong and issues tab is also not working properly.

This PR should fix that.


halkeye:
ouch. I knew this was a bigger change and why i kept putting it off.

Thanks for catching it.

halkeye (approved):

halkeye:
feel free to merge (or auto merge) once ci passes

Issue #792, opened 2021-08-31
Include excerpt and content in search index


zbynek:
Currently only plugin ID, title and developer names/is are searchable:
https://github.com/jenkins-infra/plugin-site/blob/master/src/utils/algolia-queries.js#L74

We should definitely add excerpt to that attribute list and I'd suggest to add content too.

@halkeye are these two attributes intentionally missing?

halkeye:
I vaguely think it didn't return useful results so we didn't keep it enabled. Do you have an example term you were thinking?

halkeye:
I think in this case except and content are probably going to be the same. Content by itself is too big for algolia to fully search so it's truncated. At least I think that's why it's truncated

zbynek:
My main concern is that there are no good results for checkstyle, spotbugs, pmd -- all of these should show the warnings-ng plugin. Currently these words only appear in the wiki content, truncating it to [5000 characters](https://github.com/jenkins-infra/plugin-site/blob/master/src/utils/algolia-queries.js#L34) should not be a problem. 

If you prefer using the excerpt (that's also what plugin manager does, since excerpt comes from the hpi file),  we should remove `content` from the data that's sent to Algolia and only send excerpt.

I'm sure there are more examples where including the excerpt would make a difference, e.g https://plugins.jenkins.io/svncompat14/ does not come up when looking for "subversion", though I doubt anyone cares about svn 1.4 anymore.

halkeye:
yea I don't have any objections to it, just vaguely remember not including
it for a reason.

This is def something that @markewaite and docs team should weigh in on.

Honestly there's no real reason not to try, We could make the change for a
day or so and see how things work out, and revert if necessary.

PR #810, opened 2021-09-19
Use excerpt for search, no wiki content in Algolia

zbynek:
Related to issue #792

Let Algolia use excerpt for indexing. Content is still not used => don't even send it to Algolia.


halkeye (approved):
lgtm

PR #831, opened 2021-10-23
[Matomo] Record which tab on plugin page was clicked

halkeye:
* Track as page nav (lets see how clear that is)
* Track as custom events (lets see how useful it is)
* Change tabs from nav to pills for more visibility

image


halkeye:
merging cause i really want to see data, we can revert if people hate the change

PR #851, opened 2021-11-25
Fix html rendering for new markdown wiki types

halkeye:
I made it more complicated than it needed to be. I was excited by the
field extension learning I had.

Created a new node content type for plugin site api, and then the body
will either be MarkdownRemark, or PluginHtml, and its easy to query,
only one will ever be returned.


halkeye:
Self approving as we have some broken pages

PR #855, opened 2021-11-29
Fixes #853 - Link directly to jira open issues

### LINKED ISSUE #853: Plugin pages - "Open issues" link shows issues that are Closed
[Plugin](https://plugins.jenkins.io) pages like https://plugins.jenkins.io/git-client/ include helpful links to "Report an issue" and "Open issues".  The "Open issues" link opens a Jira query that also includes issues that are closed. 

Could the "Open issues" link be extended to exclude issues that are closed?

## Screenshots 

screencapture-plugins-jenkins-io-git-client-2021-11-28-06_34_51-edit

## Possible Solution 

Change the link from https://issues.jenkins.io/issues/?jql=component=17423 to https://issues.jenkins.io/issues/?jql=component%3D17423%20and%20status!%3DClosed

Several different Jira workflows are used by Jenkins plugins.  Some do not use `Closed` as the final state.  Even with that, I believe that requesting `status!=Closed` is a safe enhancement to the "Open issues" link.

halkeye:
* Refactor issue tracker component so its a component
* IssueTrackerComponent shows all types
* IssueTrackerComponent links to jira open issues

Test Server: http://next-cherries.surge.sh/scp/

Screenshot:
image


halkeye:
added test server and screenshot. Not 100% sold on having both types of issue trackers there.

MarkEWaite:
For those plugins that support both GitHub and Jira issues, the two issue links seem correct to me.  I didn't see two issue links on the test site when I looked at the git plugin, so I assume that those who have not enabled GitHub issues will not have an issue link for GitHub issues.

halkeye:
> so I assume that those who have not enabled GitHub issues will not have an issue link for GitHub issues.

yea, whatever the metadata/json/updatecenter/etc says is shown.

i just its a lot of text and doesn't really make sense to someone just wanting to report a bug

maybe issueTrackers[0].reportUrl, then all viewUrls? donno

timja:
> > so I assume that those who have not enabled GitHub issues will not have an issue link for GitHub issues.
> 
> yea, whatever the metadata/json/updatecenter/etc says is shown.
> 
> i just its a lot of text and doesn't really make sense to someone just wanting to report a bug
> 
> maybe issueTrackers[0].reportUrl, then all viewUrls? donno

I doubt that anyone who has GitHub configured wants Jira as well reported, they probably just haven't bothered to migrate / shut down the old tracker.

but seems fine anyway

halkeye:
> I doubt that anyone who has GitHub configured wants Jira as well reported, they probably just haven't bothered to migrate / shut down the old tracker.

That actually can be noted In RPU (see first code block https://github.com/jenkins-infra/repository-permissions-updater/blob/master/README.md#declaring-issue-trackers). I agree though, splitting reporting isn't a good plan which is why after thinking about it I proposed only showing report for the first one in the list.

Buuut now re-reading that doc, I think following exactly what the yaml has is the right solution. I'll merge tomorrow unless I hear objections.

PR #880, opened 2021-12-02
 Autolink headers (html version)

halkeye:
* forked transformer-rehype which is no longer maintained
* created gatsby-rehype-autolink-headers plugin, based on remark
  (markdown) plugin

Eventually i'd like to move to yarn workspaces and a mono repo so all
the required packages can be in this repo

Demo site:
https://next-cherries.surge.sh/azure-credentials/#using-azure-credentials-in-your-own-jenkins-plugin

Debating if we should prefix the ids at all to prevent collisions with anything build in. Github does user-content-${id} for this reason.

### DIFF
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -64,7 +64,7 @@ module.exports.plugins = [
                 {
                     resolve: '@halkeye/gatsby-rehype-autolink-headers',
                     options: {
-                        elements: ['h1', 'h2', 'h3'],
+                        prependId: 'plugin-content-',
                         isIconAfterHeader: true
                     }
                 },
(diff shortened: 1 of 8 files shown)


timja (approved):
it works!

One possible improvement for later is there seems to be some artifacts on the line of the svg.

image

halkeye:
I have trouble seeing graphical things even when things are pointed out. It's not hard to replace the SVG though.

I'll merge tomorrow night unless objections. I'm still leaning towards prefixing ids.

halkeye:
okay, next steps are to try and figure out how to get the markdown and the HTML (API) versions to match up.

PR #881, opened 2021-12-05
Add tests

halkeye:
Starting to add more tests so its safer to merge/automerge dependencies. And just cause tests++

### DIFF
--- a/src/components/LineChart.jsx
+++ b/src/components/LineChart.jsx
@@ -1,8 +1,21 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import {Line} from 'react-chartjs-2';
-import moment from 'moment';
 import {ResizeObserver as ResizeObserverPolyfill} from '@juggle/resize-observer';
+const MONTHS = {
+    0: 'Jan',
+    1: 'Feb',
+    2: 'Mar',
+    3: 'Apr',
+    4: 'May',
+    5: 'Jun',
+    6: 'Jul',
+    7: 'Aug',
+    8: 'Sep',
+    9: 'Oct',
+    10: 'Nov',
+    11: 'Dec',
+};
 
 const calculateMax = (data) => {
     return 1.2 * Math.max(...data);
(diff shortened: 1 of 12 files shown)


src/components/LineChart.test.jsx
```diff
@@ -0,0 +1,33 @@
...
+            {'timestamp': 1617235200000, 'total': 5079},
+            {'timestamp': 1619827200000, 'total': 4946},
+            {'timestamp': 1622505600000, 'total': 4960},
+            {'timestamp': 1625097600000, 'total': 4837},
+            {'timestamp': 1627776000000, 'total': 4827},
+            {'timestamp': 1630454400000, 'total': 4726},
+            {'timestamp': 1633046400000, 'total': 4647},
+            {'timestamp': 1635724800000, 'total': 4704}
+        ];
+        const {getByRole} = render(<LineChart installations={installations} />);
```
halkeye, line 27: sadly this doesn't actually catch the esm related bugs upgrading to chartsv4. But its still good.

src/components/PluginLastReleased.test.jsx
```diff
@@ -0,0 +1,39 @@
...
+        const {container} = render(<PluginLastReleased releaseTimestamp="2017-02-09T15:19:10.00Z" buildDate="Feb 09, 2017" />);
+        // screen.debug();
+
+        expect(container).toHaveTextContent('Released: 5 years ago');
+        expect(container.querySelector('time')).toHaveAttribute('datetime', '2017-02-09T15:19:10.000Z');
+        expect(container.querySelector('time')).toHaveAttribute('title', '2017-02-09 15:19');
+    });
+    it('only timestamp and no buildDate', async () => {
+        const {container} = render(<PluginLastReleased releaseTimestamp="2017-02-09T15:19:10.00Z" />);
+        // screen.debug();
```
zbynek:
```suggestion
```

src/components/PluginLastReleased.test.jsx
```diff
@@ -0,0 +1,39 @@
...
+
+        expect(container).toHaveTextContent('Released: 5 years ago');
+        expect(container.querySelector('time')).toHaveAttribute('datetime', '2017-02-09T15:19:10.000Z');
+        expect(container.querySelector('time')).toHaveAttribute('title', '2017-02-09 15:19');
+    });
+    it('no timestamp and only buildDate', async () => {
+        const {container} = render(<PluginLastReleased buildDate="Feb 09, 2017" />);
+        // screen.debug();
```
zbynek:
```suggestion
```

src/components/LineChart.jsx
```diff
@@ -1,8 +1,21 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import {Line} from 'react-chartjs-2';
-import moment from 'moment';
 import {ResizeObserver as ResizeObserverPolyfill} from '@juggle/resize-observer';
+const MONTHS = {
```
zbynek, line 5: we still use `moment` in other places, why not keep it here and hardcode month names instead?

halkeye: Moment is only used in one other place and it's relatively trivial to get rid of

halkeye: It's a huge bundle for the front end when it's not even needed. I think I can force graphql to make everything timestamps

PR #889, opened 2021-12-07
Fix Renovate's automerge config

zbynek:
Assuming the goal was to automerge all dev dependencies and all non-major production dependencies, this should be the correct syntax, as far as I understand https://docs.renovatebot.com/configuration-options/#matchdeptypes


halkeye:
oh nice. Probably should fix up package.json a bit so all the gatsby* and react* packages are not dev deps.

halkeye (approved):

zbynek:
@halkeye we could also just drop the special case for dev dependencies here and always approve major updates manually, having Gatsby as dev dependency makes sense semantically.

halkeye:
yea i've always struggered with dev vs not dev in gatsy projects.

I think your right, gatsby should be dev, but react-charts-js should absolutely be dep.

Your call, I'm happy leaving as is, but not having to test every minor gatsby thing would be nice too

halkeye:
re labels nice find!

PR #907, opened 2021-12-14
Move everything to a mono repo

halkeye:
This allows us to make new rehype (HTML parser) plugins that have independent testing/dependancies/etc. Eventually moving to esm.


halkeye:
@zbynek as you are a major contibutor i'd like your opinions on this. I'd like to do it, but not at a cost of ease of use.

timja (approved):

zbynek:
It's good to have the rehype plugin in the same repo. Why is `plugin-site` in `plugins` folder? Why switch (back) to yarn?

halkeye:
Npm workspaces don't really work, yarn 3 workspaces are amazing.

The Gatsby local plugins as they were, were not really packages, they didn't have thier own dependancies, kinda but not really had thier own eslint or babel specifics.

This makes everything its own official full package with dependancies, configs, etc. 

Down the road I'm hoping i can have the plugins upgrading to using latest rehype/remark/p-queue versions which are ESM only, while the plugin site itself can be a hybrid with the ESM wrapper. As it was I couldn't get babel, eslint, Gatsby and jest to all work nicely at the same time with ESM. I'd fix one then lose the others.

Being able to node -r esm Gatsby build, and having the plugin site code all babeled, while having plugins fully tested in without needing babel would fix it.

I also want to pull in https://github.com/halkeye/gatsby-plugins/tree/master/packages so all the modules are in one place, and could be released under Jenkinsci and not me.. Maybe

I'm typing this out on my phone so hopefully it makes sense?

halkeye:
i could rename plugins to packages for a bit of consistency with other mono repos

halkeye:
@timja how's your git foo, whats the best way to resolve this conflict?

halkeye:
okay, just merged master then overwrote package.json

timja:
For anything crazy I just git checkout main filename and reapply changes :p and then regenerate yarn lock file

PR #928, opened 2021-12-18
Fix Babel config

halkeye:
For some reason during the yarn migration I somehow configured babel to
not transpile commonjs or something? React.createElement seemed to be
undefined when doing the import statements.

* Used the latest babel.config.json suggested by gatsby
* Moved it to root so it could be re-used
* Told eslint to use it


halkeye:
if this works,  https://github.com/jenkins-infra/plugin-site/pull/927 can probably be un-reverted. Re-reverted? re-comitted?

PR #937, opened 2021-12-18
Fix #914 by flattening out version and sorting by that

### LINKED ISSUE #914: Incorrectly sorted releases
https://plugins.jenkins.io/workflow-cps-global-lib/#releases currently lists 544.vff04fa68714d above 545.v7b28cce323cf even though it is older. Please make sure you are using Maven's version parser, https://github.com/jenkinsci/lib-version-number, or something compatible.

halkeye:
Originally it was sorting by buildDate, but that only returns the date, not time
Switch over to using `@machineVersion` gatsby field extension, which
turns 1.2.3 to 00001_00002_00003 which makes it way easier to sort.

Create a new module so it can be tested and eventually added to
gatsby-jenkins.io

### DIFF
--- a/plugins/gatsby-jenkinsci-fieldextensions/.eslintrc.js
+++ b/plugins/gatsby-jenkinsci-fieldextensions/.eslintrc.js
@@ -1,0 +1,109 @@
+module.exports = {
+    'env': {
+        'browser': true,
+        'es6': true,
+        'node': true
+    },
+    'extends': [
+        'eslint:recommended',
+        'plugin:import/recommended',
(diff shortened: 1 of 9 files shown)


jglick:
> turns 1.2.3 to 00001_00002_00003

Not quite what https://github.com/jenkinsci/lib-version-number would do but sounds close enough for most cases. Thanks!

halkeye:
yea I can't actually do custom logic in the sorters, it only supports alphanumeric string sorter. I'll probably open up an issue with gatsby at some point. But this will work for a while yet

PR #941, opened 2021-12-20
Fix canonical url so its not jenkins.io/template

halkeye:


timja:
Why you need approval to deploy?

timja (approved):

halkeye:
> You’ve received a Deploy Preview request from a fork by an unrecognized author

> This site is using the default sensitive variable policy, which requires site member approval for untrusted deploys when potentially sensitive environment variables are detected in your site settings. You can change your sensitive variable policy.

I didn't know that was a setting. We don't have any sensitive variables though, so I may try and disable it, but also its one click per author, so /shrug

PR #950, opened 2021-12-24
Fix font awesome and add report icon like jenkins.io

halkeye:
Trying to load font files remotely without the right headers will cause
cors errors. So lets pull it into plugin site and use it directly


timja (approved):

halkeye:
wait how did github actions know its a bug? where was it configured?

timja:
> wait how did github actions know its a bug? where was it configured?

release drafter by the looks of it:
https://github.com/release-drafter/release-drafter#autolabeler

https://github.com/jenkinsci/.github/blob/f1a34987f2919f039282a13b8741e3c463d18bc6/.github/release-drafter.yml#L90-L94

zbynek:
Maybe font awesome is not needed at all, if the only character used is the warning sign and we already have the warning sign as an SVG in this repo.

halkeye:
oh yea, thats totally another, and possibly better option. I was looking for an easy win to try out deployments.

That being said, the loading of font awesome currently doesn't work. Maybe still exclude it, but use the SVG instead?

zbynek:
> Maybe still exclude it, but use the SVG instead?

Sounds good!

zbynek:
Replaced by https://github.com/jenkins-infra/plugin-site/pull/1152

PR #961, opened 2022-01-01
Remove all the backend usage of plugin site API

halkeye:
* Removed the root .eslintrc which was breaking things
* Still using api for releases (though have a couple ideas)
* Try to render README.md, readme.md, README.adoc, readme.adoc,
  whichever comes back first


timja (approved):

halkeye:
whoops,  jenkinsfile isn't updated to node 16, tomorrow task I guess

plugins/gatsby-source-jenkinsplugins/utils.js
```diff
@@ -124,33 +124,36 @@ const getPluginContent = async ({wiki, pluginName, reporter, createNode, createC
...
+        if (wiki.url.startsWith('https://github.com/') || wiki.url.startsWith('https://www.github.com/')) {
+            const promises = [
+                {filename: 'README.md', type: 'text/markdown'},
+                {filename: 'readme.md', type: 'text/markdown'},
+                {filename: 'README.adoc', type: 'text/asciidoctor'},
+                {filename: 'readme.adoc', type: 'text/asciidoctor'},
+            ].map(async type => {
+                try {
+                    let url = wiki.url.replace('https://github.com/', 'https://raw.githubusercontent.com/');
+                    if (!url.includes('/tree/')) {
```
zbynek: This `if` may need an `else`

halkeye: can you give me an example of why?

its passing on https://ci.jenkins.io/blue/rest/organizations/jenkins/pipelines/Infra/pipelines/plugin-site/branches/PR-961/runs/4/nodes/33/steps/39/log/?start=0

the only one that triggers "has no content that can be fetched from" that is github related that I can see is "error DotCi-Fig-template has no content that can be fetched from http://github.com/DotCi/DotCi-Fig-templat"

I nudged ulli on slack asking if he knew how to setup a warnings-ng report for these lines.

halkeye: btw, anything i saw with tree/ in it, already had the branch/ref
```
 ☎  ~  %   curl -qLs 'https://updates.jenkins.io/current/plugin-documentation-urls.json' | jq '.' | grep tree -B 1
  },
  "absint-astree": {
--
  "cloudify": {
    "url": "https://github.com/jenkinsci/cloudify-plugin/tree/cloudify-1.0.8"
--
  "ibm-application-security": {
    "url": "https://github.com/jenkinsci/appscan-plugin/tree/deprecatedPlugin"
--
  "kryptowire": {
    "url": "https://github.com/jenkinsci/kryptowire-plugin/tree/master/docs"
```

zbynek: I didn't test it locally, but I think the string `tree/` should be removed from the URL. The log you linked says it's downloading from https://raw.githubusercontent.com/jenkinsci/cloudify-plugin/tree/cloudify-1.0.8/README.md , should be just https://raw.githubusercontent.com/jenkinsci/cloudify-plugin/cloudify-1.0.8/README.md (failed fetch is currently not logged, AFAIU).

halkeye: oh good catch!

> failed fetch is currently not logged, AFAIU).

failed as in 404? I specifically disabled that because 3/4 of those URLs would be 404 and that's a lot of noise. I'll add that plugin to my specific tests and try again.

plugins/gatsby-source-jenkinsplugins/utils.js
```diff
@@ -104,13 +116,46 @@ const getPluginContent = async ({wiki, pluginName, reporter, createNode, createC
...
-        data.wiki.content = $.html();
+        if (ASCIIDOC_BLOB_RE.exec(wiki.url)) {
+            const matches = wiki.url.match(ASCIIDOC_BLOB_RE);
+            const url = `https://raw.githubusercontent.com/${matches[1]}/${matches[2]}/${matches[3]}/${matches[4]}`;
+            const body = await requestGET({reporter, url: url});
+            if (body) {
+                return createWikiNode('text/asciidoctor', url, body);
+            }
+        }
+        if (wiki.url.startsWith('https://github.com/') || wiki.url.startsWith('https://www.github.com/')) {
```
zbynek: www.github.com is never used, http://github.com is (just 4 plugins though, also might be better to normalize in update center and not worry about it here)

plugins/gatsby-source-jenkinsplugins/utils.js
zbynek: maybe only log something if all 4 attempts end up as 404? (Non-blocking either way.)

halkeye: thats what I tried to do, I should abstract it into a function so I can test it more.

plugins/gatsby-source-jenkinsplugins/utils.js
halkeye: good call, i'll switch it to `new URL('http://github.com').host === 'github.com'` and seperately make an upstream PR to handle http://

halkeye: fixed thanks

plugins/gatsby-source-jenkinsplugins/utils.js
halkeye: fixed, but i havn't had time to fully locally test it (been busy helping infra stabilize), so gonna see what CI does for a full build

zbynek (changes requested):
All images in git plugin docs disappeared because of a 400 http status (wrong base folder for resolving image filenames). Also IIRC we can't load SVG images via raw.githubusercontent, we have to keep using https://cdn.jsdelivr.net/ (maybe for all images).

halkeye:
> All images in git plugin docs disappeared because of a 400 http status

image

So not all, went to look at one of the disappearing images on the git plugin

image

looks like github is more forgiving about bad URLs than we are. I'll think about a solution, could just ignore the fact that its a absolute URL and force it to be relative, but also kinda okay with it breaking due to bad data.

halkeye:
I honestly don't remember what was left undone. Everything seems to work though.

I remember we had some absolute vs relative path issues for https://deploy-preview-961--jenkins-plugin-site-pr.netlify.app/git/ but everything seems good.

If this gets merged, we can then kill off plugin-site-api, and move everything to netlify and not need to do the azure deploy stuff

halkeye:
@zbynek any chance you could review this?

zbynek:
@halkeye the absolute image URL problem is still there:
https://deploy-preview-961--jenkins-plugin-site-pr.netlify.app/slack/ -- I think it's better to fix that in plugin site than create PRs for all the plugins, so that maintainers can test their readme changes on GitHub and get them renered correctly on plugins.jenkins.io

for SVGs I get no error in console, but they seem to be simply missing from DOM:

* https://deploy-preview-961--jenkins-plugin-site-pr.netlify.app/configuration-as-code/ 
* https://deploy-preview-961--jenkins-plugin-site-pr.netlify.app/design-library/

In general it seems a bit wasteful to try all the readme URLs for all plugins each time the site is rebuilt (for CI builds that run more than once per hour it may run into the 5000 requests/hour limit mentioned in https://github.com/github/docs/issues/8031#issuecomment-881427112 ), but I can't think of a better solution.

halkeye:
> but I can't think of a better solution.

we could call the API to get the readme data directly, but i don't think its better, and would add requireing github tokens to the build.

I'll fix up the images. I couldn't find new broken stuff, but I'll admit i didn't spend a lot of time digging into it.

halkeye:
> * https://deploy-preview-961--jenkins-plugin-site-pr.netlify.app/configuration-as-code/

image

this is going to be interesting. I don't think we want to blanket support html. Lemme see how I can make it safe parsing html.

PR #972, opened 2022-01-06
Use latest builder image with yarn properly setup

halkeye:
See https://github.com/jenkins-infra/helpdesk/issues/4
and https://github.com/jenkins-infra/docker-builder/pull/34

### DIFF
--- a/Jenkinsfile_k8s
+++ b/Jenkinsfile_k8s
@@ -81,7 +81,7 @@ spec:
   }
 
   stages {
-    stage('NPM Install') {
+    stage('Install Dependancies') {
       steps {
         sh 'NODE_ENV=development yarn install'
       }
--- a/Jenkinsfile_k8s
+++ b/Jenkinsfile_k8s
@@ -53,8 +53,7 @@ spec:
   automountServiceAccountToken: false
   containers:
     - name: "jnlp"
-      image: "jenkinsciinfra/builder"
-      imagePullPolicy: Always
+      image: "jenkinsciinfra/builder:2.0.3"
       resources:
         limits: {}
         requests:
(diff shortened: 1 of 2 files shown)


Jenkinsfile_k8s
```diff
@@ -84,16 +81,6 @@ spec:
...
-      steps {
-        sh '''
-          mkdir "${HOME}/.npm-packages"
-          npm config set prefix "${HOME}/.npm-packages"
-          npm install --global yarn
-        '''
-      }
-    }
-
     stage('NPM Install') {
```
timja: nit outdated from other PRs where it switched to yarn

timja (approved):

renovate.json
```diff
@@ -26,6 +26,12 @@
         }
     ],
     "packageRules": [
+        {
+            "matchPackageNames": [
+                "jenkinsciinfra/builder"
```
timja, line 31: is renovate going to be able to update that? we'll probably remove the builder from the repo at some point anyway and make it the default for infra.ci so it doesn't need to be here

Jenkinsfile_k8s
halkeye: yea thats fair, it was kinda intentional. I felt NPM install was clearer than yarn, but I've switched it to "install dependencies"

renovate.json
halkeye: it *Should* based on https://github.com/jenkins-infra/plugin-site/blob/master/renovate.json#L14-L27 but i haven't yet been able to figure out how to get it to test the configs.

timja: o that's cool

timja (approved):

PR #973, opened 2022-01-06
Use full sha in the Jenkinsfile so renovate regex matches it

halkeye:
tested via https://regex101.com/

I'll do another PR later for the nodejs versions in Jenkinsfile, since #961 updates node anyways


timja (approved):
if it works

halkeye:
> if it works

we'll see, its fun to learn :)

PR #977, opened 2022-01-06
Have one github preview environment per PR

halkeye:
Right now when one PR deploys, the previous deploy gets marked as inactive since they are the same env. So make env separate for each PR.

### DIFF
--- a/Jenkinsfile_k8s
+++ b/Jenkinsfile_k8s
@@ -7,7 +7,7 @@ def recordDeployment(owner, repo, ref, status, environmentURL, environment = "pr
         "required_contexts": [],
         "auto_merge": false,
         "auto_inactive": false,
-        "transient_environment": false,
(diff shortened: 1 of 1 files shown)


timja (commented):
We may end up with a million environments in the GitHub ui with this though? Isn’t transient what we want for this?

halkeye:
I really don't know. I went and took a look at https://github.com/nwtgck/actions-netlify/deployments and they are doing the exact same thing.

> Specifies if the given environment is specific to the deployment and will no longer exist at some point in the future.

I guess I'm confused as to this exact sentence. I took it as in this exact behavior, its a temporary link that will go away. But in our case its a link specific to this deployment that won't go away.  We can certainly try it instead, I don't like the many different environments thing.

timja (approved):

PR #1059, opened 2022-02-15
Describe plugin installation options

zbynek:
* Add a button that shows descriptions of installation options for the latest version. The button replaces the plugin ID which can be copied from elsewhere (address bar, dependencies tab, the installation options dialog).
* Move the CLI and direct download options together in the releases tab (below the release notes)
* Do not show the checksums in the UI, provide a link for the few interested users instead. Making the checksums collapsible part of the release card instead would also be an option, but this way it's again possible to reach the update site listing that @daniel-beck mentioned was missing.

image

CC @daniel-beck  any feedback on the wording would be appreciated

Fixes https://issues.jenkins.io/browse/WEBSITE-233


halkeye:
Looks like the tests are failing.

It sounds like a decent idea. I'm not sure the gray button shoving everything down is the best plan.  Maybe to the right of the center column, instead of right of entire page?

old screenshot for comparison.

image

plugins/plugin-site/src/components/InstallInstructions.jsx
```diff
@@ -0,0 +1,46 @@
...
+            <ModalBody>
+                <ol>
+                    <li>
+                        <a href="https://www.jenkins.io/doc/book/managing/plugins/#from-the-web-ui">Using the GUI:</a>
+                        {' From your Jenkins dashboard navigate to '}
+                        <strong>Manage Jenkins &gt; Manage Plugins</strong>
+                        {' and select the '}
+                        <strong>Available</strong>
+                        {' tab. Locate this plugin by searching for '}
+                        <em>{pluginId}</em>
```
halkeye, line 19: this would def be something that my "my.jenkins.io" project I never finished would be good for.

zbynek:
> Looks like the tests are failing.

Thanks, fixed

> gray button shoving everything down

Better sugggestions for color welcome, this is the default for seconday buttons (should not be primary button IMHO). For the position -- it's vertically on the same level as the plugin name, so it should not take any extra vertical space. Do you think we should put it next to the tab headings?

plugins/plugin-site/src/components/InstallInstructions.jsx
```diff
@@ -0,0 +1,46 @@
...
+                        {'.'}
+                    </li>
+                    <li><InstallViaCLI pluginId={pluginId} version={pluginVersion} /></li>
+                    <li>
+                        <a href="https://www.jenkins.io/doc/book/managing/plugins/#advanced-installation">
+                            {'Using direct upload'}
+                        </a>
+                        {'. Download one of the '}
+                        <a href="#releases" onClick={e=>{toggleShowInstructions(e);location.href='#releases';}}>
+                            {'older releases'}
```
timja: doesn't need to be older and I think this fits both cases
```suggestion
                            {'releases'}
```

daniel-beck:
> Fixes https://issues.jenkins.io/browse/WEBSITE-233

Oh wow 😄

timja:
Could've sworn I posted this before but mustn't have submitted:

JEP-229 versions are overflowing atm:

image

halkeye (approved):
Oh man I forgot to come back to this. 

It looks good to me. Tim's comments should get addressed eventually, but I also think its a win.

future - I'm also feeling like it doubles up the info in the releases tab, do we want to add the button there (maybe as well) for each version?
future - Also as I'm trying to have a versioned plugins.txt I'm thinking maybe its worth adding lines for plugins.txt and plugins.yaml

zbynek:
I fixed the problem with long version numbers not shown properly reported by @timja . For the button position, I'm not sure whether it's a good idea to move it. It's using the same spot as the plugin ID was using before. Maybe this can be merged as is and if someone has ideas for layout improvement those could go to a new PR(?)

halkeye:
Yep yep. Merge when your happy and iterate

zbynek:
@daniel-beck ok to merge?

halkeye:
Imma gonna merge. I think its an improvement, we can come back afterwards and refine.

PR #1071, opened 2022-02-21
Rebuild yarn.lock

zbynek:
Update dependencies not mentioned in package.json

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -5,15 +5,6 @@ __metadata:
   version: 5
   cacheKey: 8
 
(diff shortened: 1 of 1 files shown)


halkeye:
Btw what problem is this trying to solve?

zbynek:
Some (all) of the dependabot warnings should go away if we update the transitive dependencies. I think Renovate only updates packages explicitly mentioned in package.json, so updating transitive dependencies has to be done manually.

halkeye:
its hard to review a lock file :)

there's a conflict, but it seems fine to me.

halkeye (approved):

Issue #1125, opened 2022-03-29
Issues with current Contributing Guide instructions to run the app


krisstern:
Today I was going through the instruction to run the app at https://github.com/jenkins-infra/plugin-site/blob/master/CONTRIBUTING.md#running-the-app. However, I ran into several issues, which require me to do the following to make things work instead:

```shell
yarn install
cd ./plugins/plugin-site
yarn install
yarn dev
```
Then open http://localhost:3000. 

Moreover, I had to manually set the values for `GATSBY_ALGOLIA_APP_ID` and `GATSBY_ALGOLIA_SEARCH_KEY` as environment variable before I could fetch the plugin records. The settings I used were as follows on a Mac:

```shell
export GATSBY_ALGOLIA_APP_ID=HF9WKP9QU1
export GATSBY_ALGOLIA_SEARCH_KEY=4ef9c8513249915cc20e3b32c450abcb
```

Would it be a good idea for me to open a PR to update the run instructions?

zbynek:
I think you shouldn't need the second `yarn install`. Maybe you're using some older version of `yarn`?

As for documenting the Algolia options, it would be great if you would add it (there's a ticket for it https://github.com/jenkins-infra/plugin-site/issues/631 )

krisstern:
Hi @zbynek Yes, I am using the classic version of `yarn`. Must be why.

Also, I would follow up on documenting the Algolia options and open a PR for the ticket #631 soon.

zbynek:
@krisstern OK to close this issue?

krisstern:
Yup

PR #1135, opened 2022-04-05
Remove non-Algolia Support and add Algolia documentation

### LINKED ISSUE #631: Remove Non Algoila support
* [ ] Document how to connect agolia for personal development
* [ ] Remove code accessing plugin-site-api endpoint

krisstern:
Fixes #631 and #1125 

- [x]  Document how to connect agolia for personal development
- [x]  Remove code accessing plugin-site-api endpoint


halkeye:
I can't really review it till tomorrow. It looks good but some suggestions

1) while the new docs are nice, using a dotenv file (already integrated)
means no windows vs Linux instructions
2) let's get more aggressive and not bother with the readme. Get rid of the
if check and always use Akhila, and default the two keys in the component.
Process.env.ALGOLIA || 'whatever the key in the readme is'

That way people won't end up with a broken site during development

krisstern:
Sure, let me add an `.env` file now.

krisstern:
@halkeye I am seeing this in a warning email though:

Screenshot 2022-04-05 at 9 51 53 PM

krisstern:
Tried getting rid of the `if` clause too, didn't work as expected as demonstrated, so had to revert.

halkeye:
> Tried getting rid of the `if` clause too, didn't work as expected as demonstrated, so had to revert.

can't give any feedback on "things didn't work"

> @halkeye I am seeing this in a warning email though:

do they provide a contact address? those keys are sent as is to algolia when you search, anyone can read those tokens. The ADMIN token is the one that is protected, it does writes.

zbynek:
> let's get more aggressive and not bother with the readme

I think it would be good to support both scenarios:
* development using the public algolia instance: would be nice if this PR just added the public values for GATSBY_ALGOLIA_APP_ID, GATSBY_ALGOLIA_SEARCH_KEY as fallbacks when the environment vars are not specified -- readme should only mention that this is the default
* custom Algolia instance: usefull for devvs who want to work on the search feature. Readme should mention all 3 variables (including GATSBY_ALGOLIA_WRITE_KEY) and that they can be set via `.env`.

krisstern:
I don't have the value for `GATSBY_ALGOLIA_WRITE_KEY` at the moment.

halkeye:
Yea write key is super private.

CONTRIBUTING.md
```diff
@@ -62,10 +62,25 @@ license your work under the terms of the [MIT License](LICENSE).
...
+GATSBY_ALGOLIA_SEARCH_KEY=4ef9c8513249915cc20e3b32c450abcb
+```
+
+However, to set the `GATSBY_ALGOLIA_APP_ID` and `GATSBY_ALGOLIA_SEARCH_KEY` as environment variables manually, for example if you are using Linux or macOC, please do the following:
+
+```shell
+export GATSBY_ALGOLIA_APP_ID=HF9WKP9QU1
+export GATSBY_ALGOLIA_SEARCH_KEY=4ef9c8513249915cc20e3b32c450abcb
+```
+If you are using a Windows computer you can set the above values by creating new environment variables under System.
```
halkeye: Since you have the above paragraph, its more clear than trying to tell people how to use linux or macos

```suggestion
```

CONTRIBUTING.md
```diff
@@ -62,10 +62,25 @@ license your work under the terms of the [MIT License](LICENSE).
 
 ## Requirements
 
-- node 12+
+- node 16+
 
 ## Run locally
 
+Supposedly the necessary environment variables have been set with a `.env` file. You can simply copy and paste the following in a `.env` file placed at the root of this directory:
```
halkeye:
```suggestion
You can simply copy and paste the following in a `.env` file placed at the root of this directory:
```

plugins/plugin-site/src/templates/search.jsx
```diff
@@ -4,7 +4,7 @@ import React from 'react';
 import querystring from 'querystring';
 import PropTypes from 'prop-types';
 import {navigate, useStaticQuery, graphql} from 'gatsby';
-import fetch from 'isomorphic-fetch';
+// import fetch from 'isomorphic-fetch';
```
halkeye: No commented out code please

halkeye:
My preference is to drop the useAlgolia variable and ifcheck entirely, and just default it to the keys. Then I don't think we need the section in the readme.

```
        const searchClient = algoliasearch(
            process.env.GATSBY_ALGOLIA_APP_ID || 'HF9WKP9QU1'
            process.env.GATSBY_ALGOLIA_SEARCH_KEY || '4ef9c8513249915cc20e3b32c450abcb'
        );
```

If your not comfortable with that (due to gitguardian warning or anything else) thats fine, we can merge after you implement the suggestions

plugins/plugin-site/src/templates/search.jsx
halkeye:
```suggestion
```

krisstern:
> My preference is to drop the useAlgolia variable and ifcheck entirely, and just default it to the keys. Then I don't think we need the section in the readme.
> 
> ```
>         const searchClient = algoliasearch(
>             process.env.GATSBY_ALGOLIA_APP_ID || 'HF9WKP9QU1'
>             process.env.GATSBY_ALGOLIA_SEARCH_KEY || '4ef9c8513249915cc20e3b32c450abcb'
>         );
> ```
> 
> If your not comfortable with that (due to gitguardian warning or anything else) thats fine, we can merge after you implement the suggestions

Yeah, I am not comfortable making the suggested changes, at least not now.

PR #1150, opened 2022-04-09
Use ID as fallback for developer name

zbynek:
Once again https://reports.jenkins.io/maintainers-info-report.json became corrupted, breaking the build of plugin website. This PR should make it more robust.

(The data seems correct again, still worth merging this I think)

### DIFF
--- a/plugins/gatsby-source-jenkinsplugins/utils.js
+++ b/plugins/gatsby-source-jenkinsplugins/utils.js
@@ -123,7 +123,7 @@ const processPlugin = ({createNode, names, stats, updateData, detachedPlugins, c
         names.push(pluginName);
         const developers = plugin.developers || [];
         const labels = plugin.labels || [];
-        developers.forEach(maint => {maint.id = maint.developerId, delete (maint.developerId);});
+        developers.forEach(convertDeveloper);
         plugin.scm = fixGitHubUrl(plugin.scm, plugin.defaultBranch || 'master');
         const pluginStats = stats[pluginName] || {installations: null};
         pluginStats.trend = computeTrend(plugin, stats, updateData.plugins);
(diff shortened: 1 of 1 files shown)


dduportal:
@zbynek could it be related to https://github.com/jenkins-infra/helpdesk/issues/2789 (thanks @lemeurherve for linking tasks!)?

We currently have the reports generated on both infra.ci (in the linked PR) and trusted.ci (actual "production") until we merge the linked PR + disable the job on trusted.

cc @daniel-beck for async. information

zbynek:
@dduportal the problem was that the JSON was prepended with one line that started with `jq`, unfortunately I didn't save the corrupted file. Though `jq` is the [tool used for generating it](https://github.com/jenkins-infra/infra-reports/blob/main/maintainers-info-report/maintainers-info-report.sh), I'm not sure how could it print that line into the output file.

dduportal:
Thanks for the fix and the pointer! I guess I'm responsible for the error though: https://github.com/jenkins-infra/infra-reports/pull/37/files#diff-008848931d3412aba076fd9e8e5953e00a8b40486e87b0b4c6992e3035dd7a61R14 

Note sure to catch "why did it print" the jq string though.

I would want to proceed on this migration and then iterate on fixing: let's see monday

NotMyFault:
Once the migration from trusted to infra is completed, are there plans to restore the original behavior to display the name of the maintainer instead of the ldap username again? This is a huge backstep in terms of recognisability.

dduportal:
I've closed https://github.com/jenkins-infra/infra-reports/pull/37 temporarly to avoid messing up the generation of reports.

Expect things to be fixed in ~4 hours (time to have a full run i trusted.ci + RPU).

Sorry for the inconvenience, but the infra team have no knowledge on "what is inside this script / Json / report " and we need help to have this working as expected.

Proposal plan for monday:

- Reopen the PR with the following changes:
  - If builds are not on the principal banch, then change the reports name to avoid overriding existing reports (wetehr they are on infra.ci or trusted)
  - rollback the "jq" change to keep using the jq 1.5 version downloaded on each actual run (instead of the image alpine jq)


I'll need help from you folks, to help me validate the infra.ci generated reports before merging PR and finish migration

lemeurherve:
We could generate the reports from infra (in another location) and on trusted simultaneously then compare their content.

daniel-beck:
> Once the migration from trusted to infra is completed, are there plans to restore the original behavior to display the name of the maintainer instead of the ldap username again? This is a huge backstep in terms of recognisability.

It's already restored.

PR #1162, opened 2022-04-17
At some point when we moved all wiki docs to export, the banner for migration broke

halkeye:
as noted on https://github.com/jenkins-infra/plugins-wiki-docs/pull/5#issuecomment-1070647496

### DIFF
--- a/plugins/plugin-site/src/templates/plugin.jsx
+++ b/plugins/plugin-site/src/templates/plugin.jsx
@@ -28,7 +28,7 @@ function shouldShowWikiUrl({url}) {
 }
 
 function shouldShowGitHubUrl({url}) {
-    return url?.startsWith('https://github.com/') && !shouldShowGitHubUrl(url);
+    return url?.startsWith('https://github.com/') && !shouldShowWikiUrl({url});
 }
 
 const PluginWikiContent = ({wiki}) => {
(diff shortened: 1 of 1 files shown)


plugins/plugin-site/src/templates/plugin.jsx
```diff
@@ -21,11 +21,14 @@ import PluginReleases from '../components/PluginReleases';
...
-    return url?.startsWith('https://wiki.jenkins-ci.org/') || url?.startsWith('https://wiki.jenkins.io/') || url?.includes('github.com/jenkins-infra/plugins-wiki-docs');
+    return url?.startsWith('https://wiki.jenkins-ci.org/') ||
+        url?.startsWith('https://wiki.jenkins.io/') ||
+        url?.includes('github.com/jenkins-infra/plugins-wiki-docs') ||
+        url?.includes('raw.githubusercontent.com/jenkins-infra/plugins-wiki-doc');
 }
 
 function shouldShowGitHubUrl({url}) {
-    return url && url.startsWith('https://github.com/');
+    return url?.startsWith('https://github.com/') && !shouldShowGitHubUrl(url);
```
zbynek: this would be an infinite recursion but the missing brackets prevent it
```suggestion
    return url?.startsWith('https://github.com/') && !shouldShowWikiUrl({url});
```

halkeye: oops, thanks, i sorta just rushed it.

PR #1205, opened 2022-06-22
Don't set dangerouslySetInnerHTML on script tags unless it has a body (fixes #1204)

### LINKED ISSUE #1204: [plugins.jenkins.io] Top sub-menus do not work
### Service(s)

plugins.jenkins.io

### Summary

When browsing https://plugins.jenkins.io/, the top menu contains sub-menus (the ones with arrow). They do not work as in jenkins.io. Only the direct links work, like "Blog" or "Download".

I suspect some JavaScript was not included or broken between the two websites.

Tested with Chrome, Edge.

⚠️ With Firefox, the behavior is weirder. You have to click once, and the second click will open the menu. While it's still a one click on the main website.

### Reproduction steps

_No response_

halkeye:
When I did a test build locally it seemed to work, we'll see what the preview site comes up with

fixes #1204

### DIFF
--- a/plugins/plugin-site/makeLayout.js
+++ b/plugins/plugin-site/makeLayout.js
@@ -124,7 +124,7 @@ async function makeReactLayout() {
             node.attribs.type = 'text/css';
             node.attribs.media = 'all';
         }
-        const attrs = Object.entries(node.attribs || {}).map(([key, val]) => {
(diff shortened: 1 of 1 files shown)


halkeye:
Preview works for me. Cc @Wadeck @zbynek

MarkEWaite (approved):
Looks good to me as well.  Thanks!

PR #1221, opened 2022-08-11
Remove report problem button from footer for plugins

### LINKED ISSUE #1199: Weird "Report a problem" behavior on plugin site
Plugin pages on plugins.jenkins.io have a link labeled "Report a problem" in the footer.

Usually on jenkins.io it is a content problem, but this one links to the plugin's issue tracker as a redundant link for reporting a bug (the main links are in the sidepanel).

Yet the link has a title/tooltip indicating it's for a content bug on the page:
> Screenshot 2022-05-09 at 21 58 37

All of this is very confusing and the link should probably just be gone.

zbynek:
Report problem should be available only nce per page: right column for plugins (report functionality or documentation problem to maintainers), footer on search page and homepage (report problem with plugin site)

Fixes #1199

### DIFF
--- a/plugins/plugin-site/src/components/ReportAProblem.jsx
+++ b/plugins/plugin-site/src/components/ReportAProblem.jsx
@@ -27,9 +27,9 @@ function ReportAProblem({reportProblemTitle, reportProblemUrl, reportProblemRela
 }
 
 ReportAProblem.propTypes = {
-    reportProblemTitle: PropTypes.string.isRequired,
-    reportProblemUrl: PropTypes.string.isRequired,
-    reportProblemRelativeSourcePath: PropTypes.string.isRequired
+    reportProblemTitle: PropTypes.string,
+    reportProblemUrl: PropTypes.string,
+    reportProblemRelativeSourcePath: PropTypes.string
 };
 
 export default ReportAProblem;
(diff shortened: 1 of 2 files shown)


plugins/plugin-site/src/components/ReportAProblem.jsx
```diff
@@ -3,10 +3,22 @@ import PropTypes from 'prop-types';
...
+    TODO: Describe the expected and actual behavior here
+
+    %23%23 Screenshots
+    TODO: Add screenshots if possible
+
+    %23%23 Possible Solution
+
+    %3C!-- If you have suggestions on a fix for the bug, please describe it here. --%3E
+
+    N/A`.replace(/\n */g, '%0A');
```
halkeye, line 17: I also like the splitting into many lines.

Did you actually change anything here? or just add new lines?

I'm trying to turn these into reusable components - https://github.com/halkeye/jenkins-io-react/blob/main/src/components/ReportAProblem.jsx#L11 (eventually)

halkeye (approved):
lgtm

zbynek:
@daniel-beck @timja do you think this makes sense? Alternative is to add some explanation of the different meanings of "reporting a problem"...

plugins/plugin-site/src/components/ReportAProblem.jsx
zbynek: Just added newlines. The URLSerchParams implementation looks nice!

daniel-beck:
Definitely makes sense. Generated pages on jenkins.io deliberately hide the "Report a problem" link as well, like https://www.jenkins.io/doc/developer/extensions/ or https://www.jenkins.io/doc/pipeline/steps/workflow-basic-steps/

PR #1222, opened 2022-08-13
Avoid Gatsby warning in development mode

zbynek:
`yarn dev` complains that `plugin.jsx` should only export the template and query, this moves a helper function to a separate file to make the warning disappear.


halkeye:
Lgtm but can you put it in hooks instead? It essentially is one.

PR #1227, opened 2022-08-29
Developer names link to search

### LINKED ISSUE #1212: Make authors link to search page
Since authors/developers are indexed in algolia, we could link authors to the search page so people could find other plugins owned by them.

eg: https://plugins.jenkins.io/ui/search?query=markewaite

Not sure how much it would be used though

zbynek:
Fixes #1212

### DIFF
--- a/plugins/plugin-site/src/components/PluginDevelopers.jsx
+++ b/plugins/plugin-site/src/components/PluginDevelopers.jsx
@@ -1,12 +1,13 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import {Link} from 'gatsby';
 
 function PluginDevelopers({developers}) {
     return (
         <>
             {developers.map((developer) => (
                 <div className="maintainer" key={developer.id}>
-                    <a href={`/ui/search?query=${developer.id}`}>{developer.name || developer.id}</a>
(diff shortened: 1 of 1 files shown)


plugins/plugin-site/src/components/PluginDevelopers.jsx
```diff
@@ -6,7 +6,7 @@ function PluginDevelopers({developers}) {
         <>
             {developers.map((developer) => (
                 <div className="maintainer" key={developer.id}>
-                    {developer.name || developer.id}
+                    <a href={`/ui/search?query=${developer.id}`}>{developer.name || developer.id}</a>
```
halkeye: should be link component. It does some magic to prevent the entire page reloading, just doing react stuff.

halkeye: I can find an example after work, but i think its gatsby-link

zbynek: thanks, fixed

halkeye (approved):
the query should probably be wrapped in encodeURIComponent
but i can't imagine developer id ever having special characters

lgtm

PR #1228, opened 2022-08-29
Add a GitHub icon with link next to the release title

lemeurherve:
Not displayed if there isn't any GitHub release.

I've tried all kind of `vertical-align` on the icon but didn't managed to get it correctly aligned with the release title, if anyone has an idea I'm all ear!

<details>
<summary>Screenshots:</summary>

image

Older ec2 release without GitHub link:

image

</details>

<details>
<summary>Original "Link plugin release versions to GitHub release instead of download link" PR body</summary>

As each release in `https://plugins.jenkins.io/<plugin-name>/#releases` already contains a download link in their block, and as there is no direct link to their corresponding GitHub release page, I'm proposing to make release version titles link to the corresponding GitHub release page instead of the plugin direct download link.

image

Ex: https://plugins.jenkins.io/github-api/#releases, clicking on "1.303-400.v35c2d8258028" would redirect to https://github.com/jenkinsci/github-api-plugin/releases/tag/1.303-400.v35c2d8258028 instead of https://updates.jenkins.io/download/plugins/github-api/1.303-400.v35c2d8258028/github-api.hpi (already in the "Installation options")

(Suggested by @dduportal)

</details>


halkeye:
so release will only be an object if it maps to a github release. Plugins not using github releases will have an empty link.
We also kinda talked about this a while ago, and while not ever coming to a conclusion, I think most who click on the link are trying to get the hpi.
I think i'd rather see an explicit github icon if there's a github release.

dduportal:
The initial idea is to have a « standard » way to reference a given plugin release with a permalink.

as for today, i do not know how to do this with the current website. 

Having a permalink either on the title or even on the header, would provide a centralized entry point for readers, wether the plugin uses gh release or not. (I would even argue that it would another incentive to switch to gh release)

lemeurherve:
@halkeye I've added a GitHub icon (if there is a GitHub release) as you suggested.

hervelemeur:
@zbynek thanks for the align fix!

zbynek (approved):

zbynek:
@dduportal any objection to merging?

halkeye (approved):
https://deploy-preview-1228--jenkins-plugin-site-pr.netlify.app/apimap/#releases lgtms

daniel-beck:
This is really useful, thank you!

PR #1229, opened 2022-08-30
Use release timestamp instead of build date

### LINKED ISSUE #1220: Releases show time always as midnight while appearing to have higher resolution than "day"
Plugin site right now:
> Screenshot 2022-08-11 at 21 41 33

Jenkins right now:
> Screenshot 2022-08-11 at 21 41 39

I was genuinely wondering whether that's the same release until I noticed that the tooltip on the plugin site says the release was at 00:00.

Unsure what the data source is, but if it only has day resolution, the human-friendly labels should just say "today", "yesterday", "X days ago" instead of "X hours ago", and the tooltip shouldn't give the time.

zbynek:
Fixes #1220

### DIFF
--- a/plugins/gatsby-source-jenkinsplugins/utils.js
+++ b/plugins/gatsby-source-jenkinsplugins/utils.js
@@ -463,7 +463,7 @@ const fetchPluginVersions = async ({createNode, reporter, firstReleases}) => {
     for (const pluginVersions of Object.values(json.plugins)) {
         for (const data of Object.values(pluginVersions)) {
             /*
-            buildDate    "May 13, 2020"
+            releaseTimestamp    "2012-10-10T17:27:42.00Z"
             dependencies    […]
             name    "42crunch-security-audit"
             requiredCore    "2.164.3"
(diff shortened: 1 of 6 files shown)


zbynek:
CC @daniel-beck

plugins/gatsby-source-jenkinsplugins/utils.js
```diff
@@ -472,13 +472,14 @@ const fetchPluginVersions = async ({createNode, reporter, firstReleases}) => {
             url    "https://updates.jenkins.io/download/plugins/42crunch-security-audit/2.1/42crunch-security-audit.hpi"
             version 2.1
             */
-            const date = parseDate(data.buildDate, 'MMMM d, yyyy', new Date(0));
```
halkeye, line 475: yay! only one last date parsing left.

halkeye:
seems good to me. Keep an eye on it after next deploy? Do you have/want sentry access (there's a lot of noise)?

zbynek:
I have Sentry access.

PR #1244, opened 2022-10-17
Ignore missing URL

zbynek:


plugins/gatsby-source-jenkinsplugins/utils.js
```diff
@@ -75,7 +75,7 @@ const getPluginContent = async ({wiki, pluginName, reporter, createNode, createC
             id: `${pluginName} <<< JenkinsPluginWiki`,
             name: pluginName,
             url: url,
-            baseHref: `${path.dirname(url)}/`,
+            baseHref: `${url && path.dirname(url)}/`,
```
halkeye: won't this output `false/`?
`path.dirname(url || '')` would work maybe?

Honestly if there's no url, then there probalby isn't need to generate a wiki object.

zbynek: > there probalby isn't need to generate a wiki object

yep, then the error message can be created in UI and we don't need to download it from API. I tried that in a new commit.

halkeye: I'm going to revert this line, there's no need for it, and it could result in a random false.

plugins/plugin-site/src/templates/plugin.jsx
```diff
@@ -181,21 +184,19 @@ function PluginPage({data: {jenkinsPlugin: plugin, reverseDependencies: reverseD
                         <h5>Maintainers</h5>
                         <PluginDevelopers developers={plugin.developers} />
                     </div>
-                    {shouldShowWikiUrl(plugin.wiki) &&
+                    {shouldShowWikiUrl(plugin.wiki || {}) &&
```
halkeye: I'm going to change it so shouldShowWikiUrl takes in a string, and then we can do plugin?.wiki?.url

PR #1249, opened 2022-11-08
Move suspended plugins banner to the top of the page

NotMyFault:
Moves the suspended plugin banner to the top of the page and uses the banner design used on the plugin's page, rather than the common blue.

### DIFF
--- a/plugins/plugin-site/src/components/SuspendedPlugins.jsx
+++ b/plugins/plugin-site/src/components/SuspendedPlugins.jsx
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 import {Link} from 'gatsby';
 
 function SuspendedPlugins({pluginIds}) {
-    return (!!pluginIds.length && <div className="alert alert-info">
+    return (!!pluginIds.length && <div className="alert alert-warning">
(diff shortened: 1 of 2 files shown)


zbynek:
In plugin's page the semantic of a warning makes sense ("hey, the specific plugin you're looking for is broken"), in search results it's just information ("FYI there are are also some broken plugins with somewhat related names"). Is there some particular reason for the proposed change?

NotMyFault:
> Is there some particular reason for the proposed change?

The intention is to highlight the suspension properly. If you are searching for plugins, a small box at the bottom of the page letting you know about the suspension is not much of use (implying someone actually scrolls to the bottom and reads the notification).

I was looking for suspended plugins earlier, but all I got were plugins with similar names but different functionalities, not really what you expect when you look for a specific plugin.

zbynek:
For most users suspended plugins should be the least important ones, because you can't install them and the website only has very limited information about them anyway (not even the name of the plugin, just ID and suspension reason).

oleg-nenashev:
I think we should keep the warning in the bottom of the search output, but modify the appearance of the search output entries instead.

PR #1255, opened 2022-12-10
Support web components on jenkins.io

halkeye:
Don't merge until https://github.com/jenkins-infra/jenkins.io/pull/5668


halkeye:
I would say immediately after this is merged, we should spend the time to move to the shared layout generator, so we don't need to try and keep them in sync (shared one currently doesn't have report issue or improve this page components, but could if we decided against web components).

zbynek:
AFAIK not needed after https://github.com/jenkins-infra/plugin-site/pull/1256 was merged

PR #1267, opened 2023-01-31
Fix Pagination Alignment

MondalSayantan:
- Remove 'row' class to prevent overflow on mobile devices
- Aligned pagination component with the rest of the body
- Verified that removing class does not have any other effect to the UI in different screens
- Basic Code formatting done

Before
image

Now
image


zbynek:
Please revert the unrelated changes (whitespace and brackets), otherwise looks good, thanks.

MondalSayantan:
Reverted all unrelated changes in the latest commit. Thanks for your review

PR #1270, opened 2023-01-31
Display artifact id on plugin page

### LINKED ISSUE #1260: Display artifact id on plugin page
Moved from https://issues.jenkins.io/browse/WEBSITE-795

Some plugin have vastly different name and artifactId. This can be confusing for users of containerized jenkins / plugin install tool. I am aware this is disclosed once clicking on "How to install", though I still find this hard to find. After all, the problem is not users would not know how to install plugins, they are surprised by the id being different.

Present the artifactId on the page itself.

lemeurherve:
~Same layout as suspended plugins ID~

Fixes https://github.com/jenkins-infra/plugin-site/issues/1260


zbynek:
as I said in the original ticket:
> I replaced the artifact ID with the "How to install" button to provide some context for less experienced users how to use such plugin ID; more experienced users can just copy it from the address bar of the browser.

Other package management sites (npm, rubygems,...) always provide some context for the ID, but I didn't manage to fit that into the page without using the current popup.
image

lemeurherve:
Thanks @zbynek 

As @olivergondza's concerns seem still valid to me but also taking in account your comments, I'm proposing to move this ID to the sidebar under the required Jenkins version, WDYT?

image

zbynek (approved):

olivergondza:
I like it, thanks for implementing it!

PR #1276, opened 2023-02-03
Enhancement: Dynamic pagination based on window width

MondalSayantan:
- Set the default value of pagesToDisplay to 5 which was already hard coded
- Implemented dynamic changing of pagesToDisplay based on window width
- If window width is less than 768px, pagesToDisplay is set to 2, otherwise it's set to 5
- Added a window resize event listener to update pagesToDisplay value dynamically
- Tested the implementation on various window widths and it's working as expected

Before
image

After Fix
image

### DIFF
--- a/plugins/plugin-site/src/components/Pages.jsx
+++ b/plugins/plugin-site/src/components/Pages.jsx
@@ -8,18 +8,19 @@ export default class Pages extends React.PureComponent {
         current: PropTypes.number.isRequired,
         pages: PropTypes.number.isRequired,
         pagesToDisplay: PropTypes.number.isRequired,
-        updatePage: PropTypes.func.isRequired
+        updatePage: PropTypes.func.isRequired,
+        marginPagesDisplayed: PropTypes.number.isRequired
     };
 
     render() {
-        const {updatePage, current, pages} = this.props;
+        const {updatePage, current, pages, marginPagesDisplayed} = this.props;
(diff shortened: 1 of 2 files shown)


MondalSayantan:
- Previous commit might cause overflow in smaller devices
- This commit fixes the issue by setting PagesToDisplay to 1 and marginPagesDisplayed to 1 in smaller mobile devices

image

MondalSayantan:
Hi @zbynek, any chance you review my PR?

zbynek:
Thanks fo the PR. Maybe the breakpoint could be 576px instead of 768?

MondalSayantan:
Thanks @zbynek. I have changed the breakpoint to 576px and tested it.

Please review my other PR also if possible #1277. Thanks!

krisstern:
I am seeing some different behavior in the preview site though, like for the width above the breakpoint 

Screenshot 2023-03-01 at 5 50 14 AM

and for below the breakpoint 

Screenshot 2023-03-01 at 5 50 30 AM

zbynek:
@krisstern on some range of screen sizes there is indeed a bit too much spacing around, but I think this is already an improvement and with the followup PR for centering it will be fine.
@MondalSayantan thanks!

krisstern:
> @krisstern on some range of screen sizes there is indeed a bit too much spacing around, but I think this is already an improvement and with the followup PR for centering it will be fine. @MondalSayantan thanks!

Sure! Not a problem @zbynek

PR #1286, opened 2023-02-12
Remove redundant .gitignore entries

zbynek:
Some of the files only exist in subfolder and are ignored by the .gitignore in plugins/plugin-site.
The public folder was ignored twice.


halkeye:
I don't think it really makes a difference but sure

PR #1296, opened 2023-03-09
fix plugin page tab urls so they get matched and marked active properly

Harsh3341:
up1
**Before Change**: This is how it currently looks, it doesn't show the current tab as of documentation.

up2
**After Change**: After making some changes it now show as the current tab as of documentation if no tabs are selected.

### DIFF
--- a/plugins/plugin-site/src/components/PluginPageLayout.jsx
+++ b/plugins/plugin-site/src/components/PluginPageLayout.jsx
@@ -31,10 +31,10 @@ function shouldShowGitHubUrl({url}) {
 
 function PluginPageLayout({plugin, children}) {
     const tabs = [
-        {id: 'documentation', to: `/${plugin.name}`, label: 'Documentation'},
-        {id: 'releases', to: `/${plugin.name}/releases`, label: 'Releases'},
-        {id: 'issues', to: `/${plugin.name}/issues`, label: 'Issues'},
-        {id: 'dependencies', to: `/${plugin.name}/dependencies`, label: 'Dependencies'},
+        {id: 'documentation', to: `/${plugin.name}/`, label: 'Documentation'},
+        {id: 'releases', to: `/${plugin.name}/releases/`, label: 'Releases'},
+        {id: 'issues', to: `/${plugin.name}/issues/`, label: 'Issues'},
+        {id: 'dependencies', to: `/${plugin.name}/dependencies/`, label: 'Dependencies'},
     ];
 
     const [isShowInstructions, setShowInstructions] = React.useState(false);
--- a/plugins/plugin-site/src/components/PluginPageTabs.jsx
+++ b/plugins/plugin-site/src/components/PluginPageTabs.jsx
@@ -3,26 +3,13 @@ import PropTypes from 'prop-types';
 
 import {Link} from 'gatsby';
 
-const PluginPageTabs = ({ tabs }) => {
-
-    let isAnyTabActive = false;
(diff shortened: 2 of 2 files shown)


halkeye:
can you give me links on where this is breaking on the site? I think its just a url matching difference. I don't think we need the hack to add the extra class.

Harsh3341:
> can you give me links on where this is breaking on the site? I think its just a url matching difference. I don't think we need the hack to add the extra class.

visit https://plugins.jenkins.io/script-security/ its not showing documentation as default and also while clicking on other tabs it make it as active but while clicking on documentation it still remains the smae

halkeye:
So I think the proper fix is actually the opposite of https://github.com/jenkins-infra/helm-charts/commit/6180206cdc3684d55913db8e9c88020a139bef6d

so add `rewrite ^/(.*)/$ /$1 permanent;` to that nginx config to strip off slashes.

Also update the gatsby config to not writeout trailing slashes - https://www.gatsbyjs.com/docs/reference/config-files/gatsby-config/#trailingslash

or flip it so it always requires trailing slash. But right now its a mix.

I can take care of it, but not till latetonight, so feel free to make any PRs

Harsh3341:
> rewrite ^/(.*)/$ /$1 permanent;

Thanks halkeye for a better approach

let me implement both changes and open a PR for it, later you can review it

halkeye:
```
diff --git a/plugins/plugin-site/src/components/PluginPageLayout.jsx b/plugins/plugin-site/src/components/PluginPageLayout.jsx
index 600a246..3b34fdc 100644
--- a/plugins/plugin-site/src/components/PluginPageLayout.jsx
+++ b/plugins/plugin-site/src/components/PluginPageLayout.jsx
@@ -31,10 +31,10 @@ function shouldShowGitHubUrl({url}) {
 
 function PluginPageLayout({plugin, children}) {
     const tabs = [
-        {id: 'documentation', to: `/${plugin.name}`, label: 'Documentation'},
-        {id: 'releases', to: `/${plugin.name}/releases`, label: 'Releases'},
-        {id: 'issues', to: `/${plugin.name}/issues`, label: 'Issues'},
-        {id: 'dependencies', to: `/${plugin.name}/dependencies`, label: 'Dependencies'},
+        {id: 'documentation', to: `/${plugin.name}/`, label: 'Documentation'},
+        {id: 'releases', to: `/${plugin.name}/releases/`, label: 'Releases'},
+        {id: 'issues', to: `/${plugin.name}/issues/`, label: 'Issues'},
+        {id: 'dependencies', to: `/${plugin.name}/dependencies/`, label: 'Dependencies'},
     ];
 
     const [isShowInstructions, setShowInstructions] = React.useState(false);
```

That seems to work locally. plugins site is often nginx, but sometimes netlify, so slashes are rough till we fully migrate. 

Do you want to make a new PR with the above code since its your PR that triggered this

Harsh3341:
I have also tested the same approach and got the result.

I'm gonna open a PR with that change

halkeye:
awesome. I think that means we can skip the nginx change too, i gotta dig into why we strip slashes on nginx in the first place.

halkeye:
its happy on netlify - https://deploy-preview-1296--jenkins-plugin-site-pr.netlify.app/script-security/

So lets merge and :pray:

halkeye:
hrm, something clientside seems to be redirecting https://plugins.jenkins.io/script-security/releases/ to https://plugins.jenkins.io/script-security/releases

PR #1298, opened 2023-03-09
more fixes toward standarizing everything with slashes

halkeye:
client side is redirecting plugins.jenkins.io/script-security/releases/ to plugins.jenkins.io/script-security/releases
I think its the createPage thingie (i think netlify plugin, when we have the netlify flag set, overrides it)
so just add slashes everywhere

I can't wait till we can fully netlify this


halkeye:
cc @zbynek @lemeurherve @Harsh3341

Issue #1245, opened 2022-10-20
Feature Request: see images in full screen on click


lemeurherve:
Currently when looking at plugin doc, the images can be too small to be readable, and when we click on them we are redirected to GitHub.
Example: https://plugins.jenkins.io/config-file-provider/

I'd like a modal (or similar) to be open on click instead, displaying the image filling the screen and dismissed with another click.

lzeee:
Hi,  @lemeurherve 

I'm Gong. I‘m preparing to apply for GSoC 2023: [Displaying of Plugin Health Scores](https://www.jenkins.io/projects/gsoc/2023/project-ideas/displaying_plugin_health_scores/) and trying to understand this project.

I tried to solve this issue in this PR: https://github.com/jenkins-infra/plugin-site/pull/1311

sridamul:
Will the documentation render in the local env? 
![Screenshot 2024-01-25 193953](https://github.com/jenkins-infra/plugin-site/assets/107404972/fa20aa7f-3155-46e0-bf62-0239a8b5d828)


If so can anyone help me to render the page locally?

sridamul:
@krisstern Can you help me with finding out why the documentation page is not rendered locally?

![Screenshot 2024-01-25 193953](https://github.com/jenkins-infra/plugin-site/assets/107404972/ed809692-1573-4e1b-9be3-158b9b28f639)

krisstern:
Hi @sridamul, 
I think maybe @halkeye would know more about it and be able to explain the reason why, but from what I recall the plugin docs are supposed to render like this locally.

sridamul:
> I think maybe @halkeye would know more about it

Thank you! @krisstern

> but from what I recall the plugin docs are supposed to render like this locally.

#1311 In this PR, @lzeee had rendered the [page](https://user-images.githubusercontent.com/21041025/224523901-f3a85ad8-ef0e-40cf-a94a-10c5ba853069.gif) locally.

krisstern:
Might have to do with this https://github.com/jenkins-infra/plugin-site/pull/1311#issuecomment-1465120929. That's why @lzeee could render everything on the page, but that may not be a recommended way.

halkeye:
https://github.com/jenkins-infra/plugin-site/blob/2cb4b668c3e8c9bd0f160abbd406a9e9200e6809/CONTRIBUTING.md?plain=1#L76

[...]

PR #1311, opened 2023-03-12
Feature: make github wiki images zoomable

lzeee:
Hi, @lemeurherve
This PR is solving this issue: https://github.com/jenkins-infra/plugin-site/issues/1245

I'm Gong. I‘m preparing to apply for GSoC 2023: [Displaying of Plugin Health Scores](https://www.jenkins.io/projects/gsoc/2023/project-ideas/displaying_plugin_health_scores/) and trying to understand this project. Since I'm new to the open-source community and the frontend, any suggestions would be appreciated :)

To make these images zoomable, I did these modifications:
1. First, I add a component named 'MakeImgZoomable'.
2. I build a component from the dangerouslySetInnerHTML from the github wiki content, using the package 'html-react-parser'
3. Then I process the component recursively to make each img zoomable, using the package 'react-medium-image-zoom'
4. At last, process the wiki content in the plugin.jsx, using MakeImgZoomable I created.

The screen recording is as follows:
plugin-site


halkeye:
Can you rebase this branch with master? its attributing a lot of master changes to you, which is causing the build not to run.

Also i'm super nervous about html parsing. is there a way to lightbox or something without react? maybe a data parameter with the full url and some sort of vanilla lightbox? 

Though now that I think about it, this will just make a thumbnail huge, not support multiple sized images, which may not be desired.

zbynek:
This should only be enabled for images where the intrinsic width of the image is smaller than the rendered width on screen (there is no point making the badges and the logo zoomable, it's only useful for the screenshots).

lzeee:
@halkeye  Thank you for your comments~

1. I will rebase the commits, these are indeed messy. I have read the contribution guide and I aggree that rebase is necessary.

2. I understand what you mean by now. Using dangerouslySetInnerHTML does have security risks. The Documentation tab is based on dangerouslySetInnerHTML. It is necessary to filter out dangerous content before using it. I learned from this [tutorial](https://www.knowledgehut.com/blog/web-development/react-dangerouslysetinnerhtml) that DOMPurify can be used to avoid side effects. I think this might be a good alternative. Although this security problem may not be completely related to the current image issue, I believe that it should be addressed. 
image

3. I'm sorry but I didn't quite understand what you mean by saying 'this will just make a thumbnail huge, not support multiple sized images, which may not be desired'. Does it mean that the function I implemented is not what the issue needs? Sorry again for my poor english :(

@zbynek 
Thank you for your comments~ I agree with the idea. I could do some filtering on the img tag here. It is meaningless to zoom in and out of images like tags, badges and logos.

PR #1312, opened 2023-03-12
Bugfix: add .gitattributes file to solve line ending problem

lzeee:
There is a small bug in the process of reading the detached plugins, which stuck the **yarn test** process. 
Filtering the lines by 'length>1' can solve this problem, and then the yarn test process can be completed.

Error info:
image

The empty lines in the [file](https://raw.githubusercontent.com/jenkinsci/jenkins/master/core/src/main/resources/jenkins/split-plugins.txt):
20023da4332bbc949f5551089368128

Data before filtering:
82d4278121d2debdda15b4dd9a82fea

Data after filtering
178071934bd37b71946ab0f196cbb39

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,209 @@
+# This is from: https://github.com/Danimoth/gitattributes/blob/master/Web.gitattributes
+
+## GITATTRIBUTES FOR WEB PROJECTS
+#
+# These settings are for any web project.
+#
+# Details per file setting:
+#   text    These files should be normalized (i.e. convert CRLF to LF).
+#   binary  These files are binary and should be left untouched.
+#
+# Note that binary is a macro for -text -diff.
+######################################################################
+
+# Auto detect
+##   Handle line endings automatically for files detected as
+##   text and leave all files detected as binary untouched.
+##   This will handle all files NOT defined below.
+*                 text=auto
+
+# Source code
+*.bash            text eol=lf
+*.bat             text eol=crlf
+*.cmd             text eol=crlf
+*.coffee          text
+*.css             text diff=css
+*.htm             text diff=html
+*.html            text diff=html
+*.inc             text
+*.ini             text
+*.js              text
+*.json            text
+*.jsx             text
+*.less            text
+*.ls              text
+*.map             text -diff
+*.od              text
+*.onlydata        text
+*.php             text diff=php
+*.pl              text
+*.ps1             text eol=crlf
+*.py              text diff=python
+*.rb              text diff=ruby
+*.sass            text
+*.scm             text
+*.scss            text diff=css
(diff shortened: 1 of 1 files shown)


halkeye:
Its hard to explain but its the same thing.
pretty much as long as length is not zero, its truthy and filter should still work.

If i'm wrong, and i very well might be, can you update util.test.js to catch the version that doesn't work, so we don't repeat it.

lzeee:
@halkeye 
I'm sorry, but do you mean that the two conditions you mentioned are the same? That is, is `.length` the same as `.length > 1`? From my understanding, the former may be more similar to `.length > 0` because an empty line itself contains a line break character, which gives it a length of 1. This could be the reason why the empty lines are passing the filter.

Also, could you please provide more details about the updates you mentioned for util.test.js? I'm not entirely sure what kind of test code I could contribute, but I would be happy to add more to that section.

halkeye:
I'll try and explain better when I'm on a computer tomorrow. 

Let's assume ... Oh wait you have row.length > 1 not > 0. Yea that's different.

I don't think we want to change the conditional as then "a" would fail it.

Can you try row.trim().length instead?

And in that directory where you made the change. There should be a utils.test.js that test a bunch of those functions. I don't know if there's anything that tests that function specifically though. I can look tomorrow.

What I would like to see is a test that fails without your change, and passes with your change. The simplistest test. I think you need to mock returnGET to return data, probably an array of strings.

lzeee:
@halkeye 
Thank you for your late-night reply. I understand what you mean now. The condition of` length>1` is indeed problematic, and it would be more reasonable to trim the string before checking its length.

As for the test code section, I understand now and will try to contribute some code. Thank you again.

zbynek:
I'm just wondering why didn't we run into this before, since yarn test is running as part of each build. Is this some `\r\n` vs `\n` problem that only happens on Windows and depends on Git settings?

lzeee:
@zbynek 
Hi, I did find this problem on Windows, I'll try it on my mac tonight

halkeye:
I was going to say it shouldn't because we are fetching from GitHub but you are right. This is a test. We probably mocked it. We probably should get .gitattributes setup so the os doesn't need to guess

lzeee:
It did pass the test on the macOS without the `>1`. It's a bit difficult for me to understand. I searched for .gitattributes and learned that there are differences in line endings for different operating systems. The test process downloaded a txt file and processed its content, which may have caused this problem.

54da0aedc353acb5b6d4fac3ef03304

lzeee:
@halkeye @zbynek

Hi, I tried to add a .gitattributes file and it solved the problem on the Windows.
I think it's the right solution.
I also rebased my commits to 1 commit.

PR #1319, opened 2023-03-23
Handle maintainer filter and tag clicks in search results

### LINKED ISSUE #1318: Links to filter by maintainers are broken
### Problem with this page

any search result page

### Expected behavior

Search results are filtered by the maintainer I click on

### Actual behavior

They're not

### Possible solution

_No response_

zbynek:
Fixes #1318 -- stopPropagation prevents both outer and inner link from being handled, change from `Link` to `navigate` avoids error in console about links nested in links. Change of `useEffect` needed for the links to actually have any effect.

### DIFF
--- a/plugins/plugin-site/src/templates/search.jsx
+++ b/plugins/plugin-site/src/templates/search.jsx
@@ -104,7 +104,7 @@ function SearchPage({location}) {
         page, setPage,
         query, setQuerySilent, clearQuery,
         setData
-    } = useFilterHooks({doSearch, setResults, categoriesMap});
+    } = useFilterHooks();
 
     const handleOnSubmit = (e) => {
         const newData = {sort, categories, labels, view, page, query};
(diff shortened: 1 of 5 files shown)


plugins/plugin-site/src/templates/search.jsx
```diff
@@ -110,18 +110,18 @@ function SearchPage({location}) {
         const newData = {sort, categories, labels, view, page, query};
         e.preventDefault();
         navigate(`/ui/search?${querystring.stringify(newData)}`);
-        doSearch(newData, setResults, categoriesMap);
     };
 
     const searchPage = 'plugins/plugin-site/src/templates/search.jsx';
 
+    // triggered on page load and when intenal <Link> clicked, e.g. for label
```
halkeye: Internal is spelt wrong

plugins/plugin-site/src/templates/search.jsx
```diff
@@ -110,18 +110,18 @@ function SearchPage({location}) {
...
     React.useEffect(() => {
         const qs = location.search.replace(/^\?/, '');
         const parsed = querystring.parse(qs);
         parsed.query = parsed.query || '';
         setData(parsed);
         doSearch(parsed, setResults, categoriesMap);
-    }, []);
+    }, [location]);
```
halkeye, line 124: Huh that makes sense I wonder why we didn't do it that way.

halkeye (approved):
Lgtm

PR #1322, opened 2023-03-29
New template with sticky header

zbynek:
Make the header stick on scroll again


halkeye:
class says "fixed-nav" on the preview site.

Issue #1324, opened 2023-04-07
#enhancment: changing the component style


Mahmoudgalalz:
![image](https://user-images.githubusercontent.com/42272376/230524923-55765f44-1ed9-4cd2-a61f-c9000eee87b2.png)

this is not clear, and it's returning too many plugins 

I recommend changing the layout to hold 4 cards and be centered when the side tab disappears

and changing the style to a bigger 

similar to this but adding a good style to it
![image](https://user-images.githubusercontent.com/42272376/230525454-47559380-f95c-4d01-a3b4-c6bdb7dcf200.png)
I will send a pr if this considered with my suggested UI

Mahmoudgalalz:
also considering to put the pagination in the bottom

MarkEWaite:
I prefer the current dense layout because it lets me see many plugins without scrolling.  When I'm seeking a plugin that matches a specific word (like GitHub or Bitbucket or Gitea) there are usually multiple matches that I need to consider.  The less dense layout that you're proposing will make less information fit on the screen for me.

Mahmoudgalalz:
@MarkEWaite what about pagination at the bottom, and making them centered? after collapsing the side nav

MarkEWaite:
I'll need to experiment with my Chromebook with its narrow vertical space before I offer an opinion on pagination at top or bottom.  I easily forget the experience navigating on a screen with limited vertical space when working at my desktop.

Mahmoudgalalz:
okay, also I have seen a new update on blog pagination
![Screenshot 2023-04-07 at 5 05 33 AM](https://user-images.githubusercontent.com/42272376/230532864-18ee8779-9927-4c33-b96f-33315c5058fa.png)
we can do that there, and we can consider putting it in the Jenkins-components as it used on both sites

krisstern:
I see the issue has been closed. But since @halkeye asked for feedback here is what I have found: 

Actually I prefer the current layout. It seems like the number of cards returned is responsive. So for my 13 inch laptop I see this: 

<img width="1440" alt="Screenshot 2023-04-07 at 12 25 48 PM" src="https://user-images.githubusercontent.com/88480540/230540910-102c5270-c359-4fdd-af4b-c8a723535b0c.png">

Maybe the only improvement I see is actually needed is to optimise also for the 13 inch version to display 5 cards per row instead of 4. I think that can be easily done with a change of either padding/margin settings.

Mahmoudgalalz:
[...]

PR #1326, opened 2023-04-07
:zap: fixing layouts of the search-results, searchBox

Mahmoudgalalz:
As mentioned that it's not recommended to make the plugins less per row.
I just modified the box alignment to be centered and 
SearchBox to be centered with the Algolia Logo

in both layouts List and grid 

the implementation was just adding a new class `add-m` which will add a margin to the left as the other way to fix this issue will require reconstructing all the body content 

image
image
image


Mahmoudgalalz:
This should close #1324

Mahmoudgalalz:
should I ignore the package-lock, and yarn.lock

to fix the build stage?
as they give me errors about deprecated packages and I fixed them using `npm audit --fix`

Harsh3341:
Have you tried to build it locally? Using `gatsby build`

I can see you made changes in the root directory of the project refer to [Contribution Guide](https://github.com/jenkins-infra/plugin-site/blob/master/CONTRIBUTING.md#contributing-guide) and move to plugins/plugin-site and run commands there

NotMyFault (changes requested):
> this is not clear, and it's returning too many plugins

> I recommend changing the layout to hold 4 cards and be centered when the side tab disappears

This PR lacks a proper motivation and clear description, which issue it aims to address.
The current layout presents up to 24 plugins on a 1920x1080 display, 50 per page, leading to the need to scroll twice to reach the page bottom. That's a compact design by choice, to display as much information within one space as possible.
That's the rationale, too. Commonly, there are multiple matches for a search term, the searcher then needs to compare them at a glance, and not by scrolling the page up and down.
Your proposal flattens out the number of displayed plugins to 4.

Could you provide your rationale justifying why reducing the page information by over 80% is an appropriate step to go, please?
I don't believe making the current layout less dense is the right step to go.
We host over 2k plugins. Your proposal makes sense if we hosted ~100 plugins only, returning a few items within a search query, but this is not the case.

Mahmoudgalalz:
Yeah, the issue was started to fix the layout, but 
@MarkEWaite mentioned that we don't need to do that

So I proposed to fix some issues related to positioning of the grids

And placing the searchBox to the center

Mahmoudgalalz:
> Have you tried to build it locally? Using `gatsby build`
> 
> I can see you made changes in the root directory of the project refer to [Contribution Guide](https://github.com/jenkins-infra/plugin-site/blob/master/CONTRIBUTING.md#contributing-guide) and move to plugins/plugin-site and run commands there

I do not change anything in root, these are the package-lock file

I will build it, and see

halkeye:
> as they give me errors about deprecated packages and I fixed them using `npm audit --fix`

don't do unrelated changes in a given pr, it makes it much harder to review.

this is a yarn based project, so running npm is not correct behavior, so now we have two lock files. Please revert yarn.lock and remove package-lock.json from this pr.

Mahmoudgalalz:
> > as they give me errors about deprecated packages and I fixed them using `npm audit --fix`
> 
> don't do unrelated changes in a given pr, it makes it much harder to review.
> 
> this is a yarn based project, so running npm is not correct behavior, so now we have two lock files. Please revert yarn.lock and remove package-lock.json from this pr.

Sorry for that, I was raising an issue, and it relates to smth else, but we discussed it there, and we found that we should leave it, and just center the content 

that what I did

and now I will revert the yarn.lock and remove the package-lock file

Thanks, @halkeye

NotMyFault:
> So I proposed to fix some issues related to positioning of the grids

> And placing the searchBox to the center

I see, your first screenshot on this PR doesn't match with the goal of the PR. Hence, my question for clarification. I would recommend removing it, if that's not going to change.

Harsh3341:
That just shows the centering of the boxes it think its ok

NotMyFault:
> That just shows the centering of the boxes it think its ok

You missed my point.

Screenshot 2023-04-08 at 21 22 44

The 1st image shows a different layout (undesired) compared to the second one (desired). If the PR does not intend to change the design matching the first screenshot, as claimed by the submitter, it doesn't belong here.

Harsh3341:
got it

Mahmoudgalalz:
> > That just shows the centering of the boxes it think its ok
> 
> You missed my point.
> 
> Screenshot 2023-04-08 at 21 22 44
> 
> The 1st image shows a different layout (undesired) compared to the second one (desired). If the PR does not intend to change the design matching the first screenshot, it doesn't belong here.

you missed smth, as recommended not to change the count of plugins, I did not change them 

I just centered the content and boxes both layouts 
in the List and grid
they weren't centered and the SearchBox wasn't too

NotMyFault:
> you missed smth, as recommended not to change the count of plugins, I did not change them

> I just centered the content and boxes both layouts

Ah I got it. This was not clear to me, given the PR description only provides very few details about the motivation. It would probably be better to provide the target in text form too, alternatively to images only.

Thanks for the explanation!

Mahmoudgalalz:
> > you missed smth, as recommended not to change the count of plugins, I did not change them
> 
> > I just centered the content and boxes both layouts
> 
> Ah I got it. This was not clear to me, given the PR description only provides very few details about the motivation. It would probably be better to provide the target in text form too, alternatively to images only.
> 
> Thanks for the explanation!

I Just edited it, Thanks @NotMyFault

NotMyFault (approved):
Thanks!

PR #1327, opened 2023-04-08
Test with Node 18

NotMyFault:
Node 14 is EOL of support since October '21, and will be EOL of security support on April 30th, 2023.

I propose we move on to node 18, the newest LTS release.


halkeye:
Previously we had trouble with getting the right version of node on the agents. I think we fixed it a while ago, so yea allowing renovate to update sounds good.

halkeye:
We do need to fix up and unify the jenkinsfiles at some point.

PR #1320, opened 2023-03-23
Displaying of plugin health score in Search Page

Harsh3341:
As "Displaying of Plugin Health Score" project idea has been withdrawn from the GSoC 2023.
So as said by @MarkEWaite we can still open a PR for a review of the implementation.

Though the project idea has been withdrawn its implementation can be useful to the developers using plugins and contributors have already researched and worked on some implementation of this idea, so together we can work on this idea.

Here is a small implementation of Displaying the Plugin Health Score on the Search Page for every plugin.

I have fetched the plugin health score data in [link](https://github.com/Harsh3341/plugin-site/blob/displaying-plugin-health-score/plugins/gatsby-source-jenkinsplugins/utils.js) and added a separate component for Plugin Health Score in [link](https://github.com/Harsh3341/plugin-site/tree/displaying-plugin-health-score/plugins/plugin-site/src/components)

Here is the deployed website - https://deluxe-pixie-154589.netlify.app/

@halkeye @zbynek  can you plz review it

Issue #1328, opened 2023-04-11
Plugin health scores not useful


daniel-beck:
### Problem with this page

https://plugins.jenkins.io/ui/search?sort=relevance&labels=&view=Tiles&page=1&query=

### Expected behavior

Provide explanations or references about the health scores that are useful to maintainers and users.

### Actual behavior

The linked page displays plugin health scores without any explanations. As a plugin maintainer who wants to learn more it's a dead end.

Additionally, the health scores are shown too prominently IMO compared to other plugin metadata, but I guess that's subjective.

Harsh3341:
Hey @daniel-beck there are further works to be done about the explanation of how healthcare is being calculated and why that particular plugin has that score 
https://github.com/jenkins-infra/plugin-site/pull/1320#issuecomment-1500861157
As mentioned previously, I'm gonna raise a PR soon for this issue

Harsh3341:
> Expected behavior
Provide explanations or references about the health scores that are useful to maintainers and users.

@halkeye is there anything else we can do for this particular issue? we can close this issue as PR #1331 solve this

halkeye:
well the comment was that the UI doesn't have really enough context or info.

I'm not sure we really fixed that at all.

halkeye:
oh maybe we did. What do you think @daniel-beck ?

daniel-beck:
Thanks for adding the link!

The health score page has the same problem (light on details, explanations, etc.) but that's a different component.

Harsh3341:
Ya, there is still not much detail about the plugin (which will be done after adding a separate component) but for this particular issue, a reference is been added to solve it.
for details, we can solve it in #1329

alecharp:
@daniel-beck please let me know on https://github.com/jenkins-infra/plugin-health-scoring how we can improve the information there. My guess is that we might have the same problem on #1329.

PR #1331, opened 2023-04-12
added info link for Plugin Health Score

Harsh3341:
https://github.com/jenkins-infra/plugin-site/pull/1320#issuecomment-1503630565

Info links for plugins is been added

### DIFF
--- a/plugins/plugin-site/src/components/PluginHealthScore.jsx
+++ b/plugins/plugin-site/src/components/PluginHealthScore.jsx
@@ -15,23 +15,24 @@ function PluginHealthScore({healthScore, name}) {
             <div>
                 <div>
                     Health Score
-                    <span
+                    <a
+                        href={`https://plugin-health.jenkins.io/scores/${name}`}
                         id={tooltipId}
-                        style={{marginLeft: '5px',
-                            color: '#0D6EFD'}}
+                        style={{marginLeft: '5px'}}
                         onClick={(e) => {
                             e.stopPropagation();
-                            setTooltipOpen(true);
+                            setTooltipOpen(!tooltipOpen);
                         }}
                     >
                         ?
-                    </span>
+                    </a>
                     <Tooltip
                         placement="top"
                         isOpen={tooltipOpen}
                         autohide={false}
                         target={tooltipId}
                         toggle={toggle}
+                        onClick={(e) => {e.stopPropagation();}}
                     >
                         View the details about plugin&apos;s
                         {' '}
--- a/plugins/plugin-site/src/components/PluginHealthScore.jsx
+++ b/plugins/plugin-site/src/components/PluginHealthScore.jsx
@@ -1,15 +1,47 @@
-import React from 'react';
+import React, {useState} from 'react';
 import PropTypes from 'prop-types';
-import {Progress} from 'reactstrap';
+import {Progress, Tooltip} from 'reactstrap';
 
-function PluginHealthScore({healthScore}) {
+function PluginHealthScore({healthScore, name}) {
     const score = healthScore.value || 0;
     const color =
     score > 80 ? 'success' : score > 60 ? 'warning' : 'danger';
+
+    const tooltipId = `tooltip-${name}`;
+    const [tooltipOpen, setTooltipOpen] = useState(false);
+    const toggle = () => setTooltipOpen(!tooltipOpen);
     return (
         <>
             <div>
-                Health Score
+                <div>
+                    Health Score
+                    <span
+                        id={tooltipId}
+                        style={{marginLeft: '5px',
+                            color: '#0D6EFD'}}
+                        onClick={(e) => {
+                            e.stopPropagation();
+                            setTooltipOpen(true);
+                        }}
+                    >
+                        ?
+                    </span>
+                    <Tooltip
+                        placement="top"
+                        isOpen={tooltipOpen}
+                        autohide={false}
+                        target={tooltipId}
+                        toggle={toggle}
+                    >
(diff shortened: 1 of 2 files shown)


Harsh3341:
Solved #1328

plugins/plugin-site/src/components/PluginHealthScore.jsx
```diff
@@ -1,15 +1,43 @@
...
+                        autohide={false}
+                        target={tooltipId}
+                        toggle={toggle}
+                    >
+                        This is the health score of the plugin. More information can be found
+                        {' '}
+                        <a
+                            href={`https://plugin-health.jenkins.io/scores/${name}`}
+                        >
+                            here
```
NotMyFault:
```suggestion
                        View the plugin's
                        {' '}
                        <a
                            href={`https://plugin-health.jenkins.io/scores/${name}`}
                        >
                            total health score.
```
Using words like 'here' for URLs isn't a good practice, see https://uxmovement.com/content/why-your-links-should-never-say-click-here/, nor do we need to state that this is the health score, given it does already say so right below the the box.

I recommend moving this to a dedicated element because clicking on the question mark links through the plugin page, rather than directly to the health score page, causing unneeded browser noise.

Harsh3341: Thanks for informing me about this good practice, and I think I should remove the link from the question mark because I already added a tooltip

NotMyFault:
Please don't mark comments as resolved when they are not. 

> I recommend moving this to a dedicated element because clicking on the question mark links through the plugin page, rather than directly to the health score page, causing unneeded browser noise.

is still to address.

I'd recommend keeping the question mark clickable, additionally to the hover.

Harsh3341:
> I'd recommend keeping the question mark clickable, additionally to the hover.

It is clickable and also can be hovered

NotMyFault:
> > I'd recommend keeping the question mark clickable, additionally to the hover.
> 
> It is clickable

Why do you think so? You removed the logic for that, after my previous comment, for no reason.

Harsh3341:
> How did you test this? On the preview site clicking the question mark does nothing.
IMG_20230413_121128

Currently I'm in my phone while clicking on "?" Tooltip is opening, also I have tested this on desktop while clicking and hover

NotMyFault:
I didn't mention the tooltip, I mentioned the linking action you previously had on the question mark. The need to hover first and click later just requires more action of the user, when there's no need for it.

Harsh3341:
> I didn't mention the tooltip, I mentioned the linking action you previously had on the question mark. The need to hover first and click later just requires more action of the user, when there's no need for it.

O.. I misunderstood what you said before, but now I get it.

What I have done is I have removed link from "?" and added `stopPropagation()` when clicked on "?" So that it does not redirect to plugin page and also I have removed link because for phone users I want toolkit to open onclick rather then directly redirecting to Health Score page. 

What needs to be done is I have to add `stopPropagation()` in tooltip `health score`

NotMyFault:
> What I have done is I have removed link from "?" and added `stopPropagation()` when clicked on "?" So that it does not redirect to plugin page and also I have removed link because for phone users I want toolkit to open onclick rather then directly redirecting to Health Score page.

Yeah, I questioned the "why". This design causes more noise for the end user, especially if the hover is not sticky, if you're not fast quick, you don't make it:

https://user-images.githubusercontent.com/13383509/231681630-e58c0faf-1a22-4cd6-a9d1-52c01c8ebee6.mp4

Harsh3341:
Noted...

With all fixes, I will push the next commit

Harsh3341:
> Yeah, I questioned the "why". This design causes more noise for the end user, especially if the hover is not sticky, if you're not fast quick, you don't make it:

just now checked the code its reactstrap tooltip [component ](https://reactstrap.github.io/?path=/docs/components-tooltip--tooltip) causing that issue of hover, tested some changes but still result remains the same, so should I remove the tooltip component and rewrite a different code for it?

NotMyFault:
> > Yeah, I questioned the "why". This design causes more noise for the end user, especially if the hover is not sticky, if you're not fast quick, you don't make it:
> 
> just now checked the code its reactstrap tooltip [component ](https://reactstrap.github.io/?path=/docs/components-tooltip--tooltip) causing that issue of hover, tested some changes but still result remains the same, so should I remove the tooltip component and rewrite a different code for it?

None of that is needed, and out of the scope of the PR. All I asked for is re-adding the ability to click the question mark.

NotMyFault (approved):
Thanks!

plugins/plugin-site/src/components/PluginHealthScore.jsx
```diff
@@ -24,6 +57,7 @@ PluginHealthScore.propTypes = {
     healthScore: PropTypes.shape({
         value: PropTypes.number,
```
halkeye, line 58:
```suggestion
        value: PropTypes.number.isRequired,
```

plugins/plugin-site/src/components/PluginHealthScore.jsx
```diff
@@ -24,6 +57,7 @@ PluginHealthScore.propTypes = {
...
     }),
+    name: PropTypes.string,
```
halkeye, line 60: you don't handle name not being provided, so it should be required
```suggestion
    name: PropTypes.string.isRequired,
```

halkeye: @Harsh3341 do you want to do a followup pr with these? I'm just going to merge.

halkeye (approved):

PR #1337, opened 2023-04-19
Fixes breaking change in plugin health scoring api

alecharp:
Follow up of #1320.

Because of (me and) https://github.com/jenkins-infra/plugin-health-scoring/pull/286, the format returned by the plugin health scoring API has changed.

This PR is to map this modification.

cc @Harsh3341 @halkeye @NotMyFault @zbynek as reviewers of the original pull request.

### DIFF
--- a/plugins/gatsby-source-jenkinsplugins/__mocks__/plugin-health-score.json
+++ b/plugins/gatsby-source-jenkinsplugins/__mocks__/plugin-health-score.json
@@ -1,22 +1,154 @@
 {
-    "aws-java-sdk-sns": {
-        "value": 96,
-        "version": "1.12.406-370.v8f993c987059",
-        "details": [
-            {
-                "key": "update-center-plugin-publication",
-                "value": 1,
-                "coefficient": 1
-            },
-            { "key": "security", "value": 1, "coefficient": 1 },
-            {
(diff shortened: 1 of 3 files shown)


halkeye:
I'm cool with this change, but questions for going forward.

1) If health service is down (or as in this case apis change), should that block the build, or should that just hide the health score because there is no valid one?
2) Okay i forgot the other question(s)

I think we are good to merge though

halkeye (approved):

Harsh3341:
This looks good🚀

This change will help in the development of the Health Score page and also now I can change the bar color on the basis of statistics as [mentioned](https://github.com/jenkins-infra/plugin-site/pull/1320#issuecomment-1503872993)

alecharp:
@halkeye I don't think we should break the site build but we should hide the missing score.

One thing I didn't think of is: scores can change and the site shouldn't have to be rebuild to see the new scores.

lemeurherve:
I was also thinking about putting in place a periodic export of plugin health scores like what's done in https://github.com/jenkins-infra/infra-reports so the plugin site could retrieve these scores from the same place instead of directly querying the plugin health score API. Then even if the plugin health score is down previous scores could be displayed.

WDYT?

PR #1343, opened 2023-04-24
Update node image for pods

NotMyFault:
We build with Node 18, therefore it makes sense to use node 18 here too.


halkeye:
todo: merge Jenkinsfile_k8s and Jenkinsfile now that we can do it.

PR #1379, opened 2023-04-24
Ban IE 11

NotMyFault:
IE 11 has been removed from our browser compatibility matrix past year, we can safely ignore it here too.


zbynek:
Did you check that this makes any actual difference?

https://browsersl.ist/#q=%3E+0.25%25+and+not+dead+and+not+IE+11

says "not IE11 already in not dead"

Issue #1393, opened 2023-04-26
Site crashes when plugin has no HealthScore


alecharp:
### Problem with this page

https://plugins.jenkins.io/ui/search?query=cdevents

### Expected behavior

We should be able to see the card of the plugin CDEvents.

### Actual behavior

The plugin `CDEvents` is a rather recent addition to the ecosystem. Because of this, the plugin has no score yet (working on this aspect). Because of this, the app crashed.
This is reproducible locally.

alecharp:
This was reported here https://github.com/jenkins-infra/helpdesk/issues/3547

alecharp:
@Harsh3341 I think we need to hide the `PluginHealthScore` for such cases.

Harsh3341:
> @Harsh3341 I think we need to hide the `PluginHealthScore` for such cases.

ya should I display the value as 0 and mention the issue in Tooltip?

> Plugin Health Score ([1](https://github.com/jenkins-infra/plugin-health-scoring)) is not seeing the plugin. This is not normal as the released occurred a few hours ago and the application should have updated its list of plugins from the update-center since. 

as mentioned 

it's better to not remove the Health Score and should mention the issue in the tooltip why its been not there

alecharp:
I'd prefer not showing a  `0`. Just mention `no score` in place of the progress bar then?

halkeye:
With https://github.com/jenkins-infra/plugin-site/pull/1394 the site no longer crashes, feel free to do a better fix

alecharp:
Thanks @halkeye for this fix. To me, this should be enough. No scores should only happen for entirely new plugins only.

alecharp:
@Harsh3341 if you want to work on a solution which would add a message saying why there is no score, you can. Otherwise, this is good enough for me.

Harsh3341:
> @Harsh3341 if you want to work on a solution which would add a message saying why there is no score, you can. Otherwise, this is good enough for me.

Currently, the fix is ok I will work on such cases later😀

krisstern:
Thank you so much for your contributions @Harsh3341!

Harsh3341:
> Thank you so much for your contributions @Harsh3341!

More work is to be done🤝, all thanks to @halkeye @alecharp @zbynek @NotMyFault for such guidance

alecharp:
[...]

PR #1394, opened 2023-04-26
If no healthscore, don't show component

halkeye:
at least a temp fix for #1393 / https://github.com/jenkins-infra/helpdesk/issues/3547


Harsh3341:
looks good, can't we add a reason for no score?

halkeye:
Fast fix first, slow fix after.

I'll merge this, make a new PR for proper behavior.

zbynek:
>  can't we add a reason for no score?

Does it ever matter? For the usage stats data can be missing for a month, for health it's just hours.

PR #1429, opened 2023-06-17
Use ES modules for gatsby config

zbynek:
This should unlock some dependency updates


plugins/plugin-site/gatsby-config.mjs
```diff
@@ -1,7 +1,12 @@
-/* eslint-disable sort-keys, no-console */
+import dotenv from 'dotenv'
+import { dirname } from 'path'
+import { fileURLToPath } from 'url'
+import algoliaQueries from './src/utils/algolia-queries.js'
```
halkeye: for sake of consistency, should this be mjs too?

halkeye:
this is so cool. I wasn't aware we upgraded gatsby enough to support this.

zbynek:
@halkeye there are still some files that we might be able to rename, but when I tried that with `ucfirst` I ran into some issues with Eslint. Also Eslint says `p-queue` is not reachable even though it clearly is. Maybe because `type:module` in `package.json` is expected by Eslint but causes problems with Gatsby? (for `plugin-site`; other workspaces are fine). Despite these I think this is ready for merge, we can improve it more when newer versions of Eslint and/or Gatsby are available.

NotMyFault (approved):
Awesome, thanks so much!

Jenkinsfile
```diff
@@ -64,6 +64,7 @@ pipeline {
     stage('Lint and Test') {
       environment {
         NODE_ENV = "development"
+        NODE_OPTIONS="--experimental-vm-modules"
```
halkeye, line 67: do we still need this for 18? 18 should have full module support
if not, do you want to goto 20 instead? I don't really like using the flag.

halkeye:
> Also Eslint says p-queue is not reachable even though it clearly is.

can you tell me how to generate that output?

zbynek:
@halkeye see https://github.com/jenkins-infra/plugin-site/pull/1429/checks?check_run_id=14690302296 -- caused by reverting Eslint config in https://github.com/jenkins-infra/plugin-site/pull/1429/commits/618e48d7d08ec4edf3a750b6ef379cc4054a8110
Apparently that's just a bug in Eslint https://github.com/import-js/eslint-plugin-import/issues/1810

Jenkinsfile
zbynek: this is only used during lint+test stage. I think it's more related to Jest implementation than Node version, see https://jestjs.io/docs/ecmascript-modules

halkeye:
I'm pretty checked out. If you think its a win, go ahead and merge it. I'm still a little uncomfortable with --experimental-vm-modules but I know how messy jest can get.

PR #1430, opened 2023-06-20
Adjusted number of plugins

StefanSpieker:
Adjusted number of plugins accoring to:
image


NotMyFault:
I'd rather keep this at 1800. That's the magic number, used in various places and documents. But I'm inclined to change it once we pass 2000 ;)

StefanSpieker:
Ok, fine for me

PR #1431, opened 2023-06-21
Display plugin health score details on individual pages

### LINKED ISSUE #1329: Plugin health scores not shown on individual plugin pages
### Problem with this page

https://plugins.jenkins.io/matrix-auth/ (etc.)

### Expected behavior

Health score (same as on the lists pages) is shown.

### Actual behavior

Health score is not shown.

### Possible solution

Individual plugin pages are a far more common landing page (e.g. from Jenkins directly) than search.

alecharp:
# Description

Resolves #1329.

This offers a new page for each plugin to display its score from plugin-health.jenkins.io.

<details>
<summary>Here is what the proposed page looks like:</summary>

Screenshot 2023-10-20 at 13-05-46 Mailer

</details>


alecharp:
For the moment, I'm struggling to get the health score fetched from the application in order to be able to display them. cc @halkeye if you have any tips / guidance I'd very much appreciate it

hervelemeur:
Reiterating a previous suggestion/question, should plugin health score data be pulled from a static export instead of being retrieved every time from plugin health score API?

alecharp:
The data are fetched when the site is built as far I understand it (https://github.com/jenkins-infra/plugin-site/blob/c9087483a1526cd9f849566e4836c6f10dd76060/plugins/gatsby-source-jenkinsplugins/gatsby-node.js#L25)

plugins/plugin-site/src/templates/plugin_healthScore.jsx
```diff
@@ -0,0 +1,41 @@
...
+
+const TemplatePluginHealthScore = ({data: {jenkinsPlugin: plugin, healthScore}}) => {
+    return (
+        <PluginPageLayout plugin={plugin}>
+            <PluginHealthScore healthScore={healthScore}/>
+        </PluginPageLayout>
+    );
+};
+
+TemplatePluginHealthScore.displayName = 'PluginPage';
```
halkeye:
```suggestion
TemplatePluginHealthScore.displayName = 'TemplatePluginHealthScore';
```

plugins/plugin-site/src/templates/plugin_healthScore.jsx
```diff
@@ -0,0 +1,41 @@
...
+};
+
+/* eslint no-undef: "off" */
+export const pageQuery = graphql`
+    query ($name: String!) {
+        jenkinsPlugin(name: {eq: $name}) {
+            ...JenkinsPluginFragment
+        }
+
+        healthScore: allJenkinsPluginHealthScore(filter: {name: {eq: $name}}) {
```
halkeye: all prefix returns an array. I think you just want `jenkinsPluginHealthScore(name: {eq: $name})`

plugins/plugin-site/src/components/PluginPageLayout.jsx
```diff
@@ -35,6 +35,7 @@ function PluginPageLayout({plugin, children}) {
         {id: 'releases', to: `/${plugin.name}/releases/`, label: 'Releases'},
         {id: 'issues', to: `/${plugin.name}/issues/`, label: 'Issues'},
         {id: 'dependencies', to: `/${plugin.name}/dependencies/`, label: 'Dependencies'},
+        {id: 'health-score', to: `/${plugin.name}/health-score/`, label: 'Health Score'},
```
halkeye: Kinda nitpicky, but lets do healthscore not health-score for a bit of consistency.

halkeye:
https://deploy-preview-1431--jenkins-plugin-site-pr.netlify.app/google-kubernetes-engine/health-score/
right now the data seems empty, is that intentional?

I kinda don't think its worth merging with a json blog, i'd rather see an iframe, or leave it alone until something a bit more presentable happens. At least a table

plugins/plugin-site/src/templates/plugin_healthScore.jsx
halkeye: also if you goto localhost:3000/__graphql you get the full graphql explorer.

halkeye: actually, jenkinsPlugin actually links them
https://github.com/jenkins-infra/plugin-site/blob/master/plugins/gatsby-source-jenkinsplugins/gatsby-node.js#L41
```
        jenkinsPlugin(name: {eq: $name}) {
            ...JenkinsPluginFragment
            healthScore: {
            }
        }
```

alecharp: thank you! This is what I didn't understand!

alecharp:
> right now the data seems empty, is that intentional?

kind off, as I didn't know where the data were. I'll be improving that now.

plugins/plugin-site/src/components/PluginPageLayout.jsx
alecharp: done in https://github.com/jenkins-infra/plugin-site/pull/1431/commits/f1280dad2e0423dbf7fd3b9109998cdd16888b2b

plugins/plugin-site/src/templates/plugin_healthScore.jsx
alecharp: I think I didn't structure the data in the API correctly because I would need to describe all the graphql for any new  `details` when this should be easily extensible. I need to change the `object` structure for an `array` so that we can easily create a table.

halkeye: the source plugin can also do some data mashing for you, and graphql can do some joins. So like if you want to do one dataset for detail definitions, and one dataset for the actual results, we can do a join.

i'm just not sure if you have a question or not. Feel free to ping me again.

plugins/gatsby-source-jenkinsplugins/utils.js
```diff
@@ -501,12 +501,13 @@ const fetchPluginVersions = async ({createNode, reporter, firstReleases}) => {
...
-    for (const pluginName of Object.keys(json.plugins)) {
-        const data = json.plugins[pluginName];
+    const url = 'http://localhost:8080/api/scores';
+    const {plugins, statistics} = await requestGET({url, reporter});
+    for (const pluginName of Object.keys(plugins)) {
+        const {value, details} = plugins[pluginName];
         createNode({
-            ...data,
+            value,
+            details,
```
alecharp: @halkeye I might need your help here. 
I don't understand why but the `details` (with https://github.com/jenkins-infra/plugin-health-scoring/pull/343, it's an array) is never put in the node.
That's the same for the `statistics` in fact. Its value is never deconstruct into the node under the graphql explorer..

plugins/gatsby-source-jenkinsplugins/utils.js
```diff
@@ -501,12 +501,13 @@ const fetchPluginVersions = async ({createNode, reporter, firstReleases}) => {
 const fetchPluginHealthScore = async ({createNode, reporter}) => {
     const sectionActivity = reporter.activityTimer('fetch plugin health score');
     sectionActivity.start();
-    const url = 'https://plugin-health.jenkins.io/api/scores';
-    const json = await requestGET({url, reporter});
-    for (const pluginName of Object.keys(json.plugins)) {
-        const data = json.plugins[pluginName];
+    const url = 'http://localhost:8080/api/scores';
```
halkeye: can you make this an env variable
```
process.env.PLUGIN_HEALTH_SCORE_API || 'https://plugin-health.jenkins.io/api/scores';
```

just so we don't break it again in the future (I was testing out the PR)

plugins/gatsby-source-jenkinsplugins/utils.js
halkeye: I'm not sure, maybe if you are changing structure maybe you have something cached? you can run `npx gatsby clean` in the plugins/plugin-site directory, just to be certain. I know for gatsby graphql schemas are inferred, so if one entry is an array and one is an object, it would get confused.
you can force graphql schema with createSchemaCustomization

you can also take a look at ./plugins/plugin-site/schema.graphql to see what the schema actually looks like, to confirm if there's anything.

Without being able to dig into the data though, I can't really do much investigation.

maybe swap out the createNode with console.log to confirm its trying to create with the right values, and nothing is null/undefined that you expect.

plugins/gatsby-source-jenkinsplugins/utils.js
alecharp: I did do a `git add -p` but it still made it way into the changest. Sorry for that.

plugins/gatsby-source-jenkinsplugins/utils.js
alecharp: I'll try the gatsby clean but when I added a key `foo` into the object I could see it and I did use `console.log` and could see the `details` array. I'll try a few more things...

halkeye: ooh, can you save your local /api/scores to a gist, and I'll just use that to poke around.

alecharp: I'll do that. Thank you for your help.

alecharp: here you go: https://gist.github.com/alecharp/4c0dab7873c98dff441ecbf5089d6577. Thank you.

However, it seems that the `npx gatsby clean` make it work. I can see the `details` array in the explorer as well as in the schema. You seems to have fixed the problem without having to reproduce it even. Nice!

alecharp:
<details>
<summary>With last commit, here is the display of the page.</summary>

Screenshot 2023-06-29 at 15-07-40 Mailer

</details>

Of course, this is still in progress, it's ugly and not easy to read.

halkeye:
> Of course, this is still in progress, it's ugly and not easy to read.

@zbynek might have some suggestions on that front.

zbynek:
@alecharp maybe you could merge the changes in `plugin-health-scoring` repo first so that the preview site build works, then it will be easier to provide feedback.

alecharp:
> @alecharp maybe you could merge the changes in `plugin-health-scoring` repo first so that the preview site build works, then it will be easier to provide feedback.

so, this might be problematic: if I merge and release the plugin-health-scoring pull request, next time the plugin-site is built, it will use the "new" format and so it won't work..

halkeye:
Can you add a version header? So you can do some sort of feature flipper based on version header?
If version == old {
return old version
}
return new version;


then later delete old version?

halkeye:
also https://pub.dev/packages/version/score seems like a nice ui to look at

zbynek:
@alecharp build stage succeeded 👍 Still some errors in CI though, I fixed the lint issue, please check the failing test.

alecharp:
<details>
<summary>Here is what the page look like with the last commit (577b554)</summary>

Screenshot 2023-10-18 at 11-40-49 Config File Provider

</details>

alecharp:
Tests are not executable locally 

```
[..]
Details:

    /***/plugin-site/plugins/gatsby-jenkinsci-fieldextensions/gatsby-node.test.mjs:5
    import {createSchemaCustomization} from './gatsby-node.mjs';
    ^^^^^^

    SyntaxError: Cannot use import statement outside a module
```

So i'm having issues to fix the test. I'm still working on it.

halkeye:
What command are you running to get that? Tests have run locally for years.

alecharp:
> What command are you running to get that? Tests have run locally for years.

I'm running `yarn test` from the root folder. I have `npm` `10.2.0`, `yarn` `3.6.3` and node `21.0.0`.

zbynek:
@alecharp sorry, the tests only work with environment variable `NODE_OPTIONS="--experimental-vm-modules"`

halkeye:
> @alecharp sorry, the tests only work with environment variable `NODE_OPTIONS="--experimental-vm-modules"`

can you use cross-env and add that to to package.json scripts test?

actually you may be able to do that in jest setup because thats pre-parsing the test files

alecharp:
That's the solution. I'm looking into putting that into the config files.

alecharp:
> actually you may be able to do that in jest setup because thats pre-parsing the test files

I couldn't use the Jest global setup file to set the environment variable. Changing the way jest is launch on the `package.json` was the solution that worked locally.

alecharp:
I'm puzzled by that last failure..

plugins/plugin-site/src/components/PluginHealthScore.jsx
```diff
@@ -1,63 +1,101 @@
...
+        weight: PropTypes.number.isRequired,
+        reasons: PropTypes.arrayOf(PropTypes.string),
+    }).isRequired
+};
+
+function ScoreDetail({data: {name, components, value}}) {
+    return (
+        <div className="pluginHealth--score-section">
+            <div className="pluginHealth--score-section--header">
+                <div className="pluginHealth--score-section--header-title">{name}</div>
```
zbynek: Would be nice to make these look more user-friendly, e.g.
```
                <div className="pluginHealth--score-section--header-title">{name.split("-").map(ucFirst).join(" ")}</div>
```
with `ucFirst` imported from [utils](https://github.com/jenkins-infra/plugin-site/blob/8f81ee3d8aecd601f4bfbab6fd2050f39db10b3d/plugins/plugin-site/src/utils/ucfirst.js#L1) would help.

plugins/plugin-site/src/components/PluginHealthScore.jsx
```diff
@@ -1,63 +1,101 @@
...
+import './PluginHealthScore.css';
 
-    const tooltipId = `tooltip-${name}`;
-    const [tooltipOpen, setTooltipOpen] = useState(false);
-    const toggle = () => setTooltipOpen(!tooltipOpen);
+function ScoreValue({value}) {
     return (
-        <>
+        <div>
+            {`${value}/100`}
```
zbynek: Maybe this could use % instead of /100? Unless you plan to have different max values for each score...

zbynek (changes requested):
Nice! I'd like to see the headings improved, my other comment about the percentages is just for consideration. 

I was also surprised to see

> No commit in the last 4 years. 0/100

for the Git plugin which has a lot of commits, but that's independent of this PR.

plugins/plugin-site/src/components/PluginHealthScore.jsx
alecharp: I don't plan to have other max value, or this would be revisited by also providing the max.

alecharp: fixed in https://github.com/jenkins-infra/plugin-site/pull/1431/commits/12148caa6fc45471deb5c88d198c6d89259d5d09

plugins/plugin-site/src/components/PluginHealthScore.jsx
alecharp: Fixed in https://github.com/jenkins-infra/plugin-site/pull/1431/commits/4c90a5e2e0381b0af4f5733a2c058dd0dc0943f7

alecharp:
> I was also surprised to see
>>    No commit in the last 4 years. 0/100
>
> for the Git plugin which has a lot of commits, but that's independent of this PR.

Yes, I have a problem there that I must fix on the application itself.

plugins/gatsby-jenkinsci-fieldextensions/package.json
```diff
@@ -6,7 +6,7 @@
   "scripts": {
     "lint": "eslint --ext .js,.jsx,.mjs .",
     "build": "exit 0",
-    "test": "jest"
+    "test": "NODE_OPTIONS=--experimental-vm-modules npx jest"
```
zbynek:
```suggestion
    "test": "node --experimental-vm-modules ../../node_modules/jest/bin/jest.js"
```

plugins/gatsby-source-jenkinsplugins/package.json
```diff
@@ -5,7 +5,7 @@
     "scripts": {
         "lint": "eslint --ext .js,.jsx,.mjs .",
         "build": "exit 0",
-        "test": "jest"
+        "test": "NODE_OPTIONS=--experimental-vm-modules npx jest"
```
zbynek:
```suggestion
        "test": "node --experimental-vm-modules ../../node_modules/jest/bin/jest.js"
```

zbynek (approved):
Looks good. One non-blocking suggestion: the health indicator on the search page should link to the health tab on the plugin site instead of https://plugin-health.jenkins.io/ and maybe the health tab could have a link to https://plugin-health.jenkins.io/ labeled "Show scoring details" or something.

alecharp:
> the health indicator on the search page should link to the health tab on the plugin site instead of https://plugin-health.jenkins.io/ 

good idea. Let me do that.

> maybe the health tab could have a link to https://plugin-health.jenkins.io/ labeled "Show scoring details" or something.

To be fair, the visualization of the health score on plugin-health.jenkins.io was not initially planned and I'm not sure we should be the two. The plugin-site should stay the source of information for plugins imo.

zbynek:
I'd like to merge this if there are no objections raised in the next 24 hours. @daniel-beck @halkeye looped

daniel-beck:
Presentation is nice 👍 It's a bit awkward that I have to click to see even the summary score, but that can always be improved later.

I suggest hiding the entry "The plugin appears in the update-center." (plus its category if otherwise empty), because that one is always 100%. Suspended plugins don't have this page, so its kinda redundant/automatic. I would keep it though is a formula or other way to see how the overall score is computed is provided in the future, where it being missing would be noticeable.

alecharp:
> It's a bit awkward that I have to click to see even the summary score

could you elaborate? There is no "hidden" section in the deployed preview so there is nothing to click on to show more details.

> I suggest hiding the entry "The plugin appears in the update-center." (plus its category if otherwise empty), because that one is always 100%.

This is coming from the fact that, some plugin were at some point in the update-center and were removed ([here](https://github.com/jenkins-infra/plugin-health-scoring/blob/main/core/src/main/java/io/jenkins/pluginhealth/scoring/probes/UpdateCenterPluginPublicationProbe.java)). The PHS application detects that and score the plugin based on that information ([here](https://github.com/jenkins-infra/plugin-health-scoring/blob/main/core/src/main/java/io/jenkins/pluginhealth/scoring/scores/UpdateCenterPublishedPluginDetectionScoring.java)).
I understand that a plugin (like `whitesource` right now) doesn't have a page on the `plugin-site` but it still exists. 

The categories we see on this pull request are the representation of the [`scoring`](https://github.com/jenkins-infra/plugin-health-scoring/tree/main/core/src/main/java/io/jenkins/pluginhealth/scoring/scores).

> I would keep it though is a formula or other way to see how the overall score is computed is provided in the future, where it being missing would be noticeable.

Sorry, I'm not sure I understand what you mean @daniel-beck. Could you elaborate?

alecharp:
> It's a bit awkward that I have to click to see even the summary score

is it that you'd like to see the global score somewhere in the right column of the plugin page?

daniel-beck:
>  I understand that a plugin (like `whitesource` right now) doesn't have a page on the plugin-site but it still exists.

I do not criticize the existence of the health check in this conversation (feedback on health checks is out of scope IMO), just it being shown here. It will never show anything other than 100% here, as otherwise there's no "here". So I suggest it be hidden.

> is it that you'd like to see the global score somewhere in the right column of the plugin page?

That, or possibly in the section header/link title. Basically, show me the summary on other pages.

Neither of these is blocking, they both can always be done later.

alecharp:
> I do not criticize the existence of the health check in this conversation (feedback on health checks is out of scope IMO), just it being shown here. It will never show anything other than 100% here, as otherwise there's no "here". So I suggest it be hidden.

I understand. I just mentioned the process because I tried not to put to much "cleverness" here, when I think it should be in the Plugin Health Scoring project. So I could hide that category from the API. As the plugin-site is the only "consumer" of that API. Would that make sense? 
However, the score is using that category, and, at this moment, it has a heavy weight.

> That, or possibly in the section header/link title. Basically, show me the summary on other pages.

I'm looking to see if I can make it happen in this pull request shortly (in the next 4hrs).

daniel-beck:
> I tried not to put to much "cleverness" here, when I think it should be in the Plugin Health Scoring project.

Good enough reason to continue showing it until users show up who are confused 👍 

> I'm looking to see if I can make it happen in this pull request shortly (in the next 4hrs).

Perhaps better as followup PR to prevent implementation choices there from blocking this? You could consider opening a second PR on top of these changes to let us already review, test, discuss that change while not blocking this one.

alecharp:
Good point. Thank you @daniel-beck for your feedback. 
Let's see if it's good enough for @halkeye as well.

PR #1492, opened 2023-09-03
Add link icon to release card header

### LINKED ISSUE #1273: Feature Request: be able to jump directly on a specific plugin release via an anchor
Currently there is no way to get a link pointing directly to a specific release.

Having something like an anchor on the release date for example would be really nice.

zbynek:
Fixes #1273 (the [remaining part](https://github.com/jenkins-infra/plugin-site/issues/1273#issuecomment-1462564927))

Also fixes some of the warnings Gatsby now produces in dev mode.

### DIFF
--- a/plugins/plugin-site/gatsby-browser.js
+++ b/plugins/plugin-site/gatsby-browser.js
@@ -61,6 +61,6 @@ exports.onClientEntry = function () {
         });
         return window.Sentry = Sentry;
     /* eslint-disable no-console */
-    }).catch(e => console.log(e));
+    }).catch(e => console.error(e));
 };
 
--- a/plugins/plugin-site/gatsby-browser.js
+++ b/plugins/plugin-site/gatsby-browser.js
@@ -9,8 +9,7 @@ exports.onClientEntry = function () {
         window.location.href = `/${window.location.pathname.split('/')[1]}/dependencies/`;
     }
 
-    require.ensure(['@sentry/browser'], (require) => {
-        const Sentry = require('@sentry/browser');
+    import('@sentry/browser').then((Sentry) => {
         Sentry.init({
             dsn: 'https://495b0bd32a5e4a2287c3fe4b061ee24f@sentry.io/1882460',
             environment: process.env.NODE_ENV,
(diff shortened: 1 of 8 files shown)


plugins/plugin-site/src/components/Pagination.jsx
```diff
@@ -7,10 +7,6 @@ function Pagination({limit, page, pages, total, setPage}) {
     const [pagesToDisplay, setPagesToDisplay] = React.useState(5);
     const [marginPagesDisplayed, setMarginPagesDisplayed] = React.useState(2);
 
-    if (total == 0) {
```
zbynek, line 10: The only place where Pagiation is called guards for `total == 0` already

NotMyFault:
Looks good to me, but the build thinks otherwise

plugins/plugin-site/src/components/PluginReleases.jsx
```diff
@@ -53,7 +53,7 @@ function PluginReleases({pluginId, versions}) {
                                         <a href={release.htmlURL} title="See the release on GitHub" className="github-icon d-flex"><ion-icon name="logo-github" /></a>
                                     </>)}
                                     <a className="anchor after" href={`#version_${version.version}`}>
-                                        <ion-icon name="link-outline"></ion-icon>
+                                        <ion-icon name="link-outline"/>
```
halkeye, line 56: I believe web components must have closing tags, not self closing.

plugins/plugin-site/gatsby-browser.js
```diff
@@ -59,7 +59,8 @@ exports.onClientEntry = function () {
                 /metrics\.itunes\.apple\.com\.edgesuite\.net\//i
             ]
         });
-        window.Sentry = Sentry;
-    });
+        return window.Sentry = Sentry;
+    /* eslint-disable no-console */
+    }).catch(e => console.log(e));
```
halkeye:
```suggestion
    }).catch(e => console.error(e));
```

plugins/plugin-site/src/components/PluginReleases.jsx
zbynek: You're right, self-closing custom elements are still not valid HTML (as far as I understand https://github.com/WICG/webcomponents/issues/624 ), but 
* ESlint prefers self-closing tags and we're using it this way for all icons in the repo
* This is .jsx so HTML parsing restrictions do not apply

NotMyFault (approved):

PR #1499, opened 2023-09-16
Show install numbers as percentages

### LINKED ISSUE #1238: Replace install count with install percentage
Install count does not generally mean anything to users as a popularity indicator unless they know what the most popular plugin is (as a benchmark for comparison) or the total install count from the stats site.

Additionally, the number of instances reporting data seems to fluctuate a bit from month to month. Whether as an artifact to stats collection or because Jenkins becomes genuinely more (or less) popular, that shouldn't affect plugin popularity -- but right now it does: A lot of plugins jumped in popularity from Feb to March this year, e.g. `matrix-auth` ~7%! In reality, as a percentage, `matrix-auth` even dropped a bit from 94.6% to 94.2%.

It would seem more useful to show install percentage.

One potential challenge with this would be that we shouldn't show raw floats/doubles with way too many digits, and 1% translates to roughly 3000 installs. To make this feasible for the long tail of plugins with less than 1% install count, there should always be two significant digits shown for install percentage:

https://plugins.jenkins.io/matrix-auth/ - 93%
https://plugins.jenkins.io/blueocean/ - 22%
https://plugins.jenkins.io/groovy/ - 9.8%
https://plugins.jenkins.io/strict-crumb-issuer/ - 2% (rounded from 2.0074%)
https://plugins.jenkins.io/solarized-theme/ - 0.36%
https://plugins.jenkins.io/generic-environment-filters/ - 0.012%
https://plugins.jenkins.io/csp/ - 0.0065%

The detailed tooltips on the graph should continue to show absolute numbers (as well as percentage), so users aren't confused why "98%" on every month isn't an exactly straight line, but the top-level label is probably better as percentage only.

Thoughts?

zbynek:
Fixes #1238

@daniel-beck one difference to your proposal: precision for big percentages is 3 significant digits (to avoid 12 identical values in graph tooltips). For small percentages I kept the proposed 2 digits (the 3rd digit might otherwise represent less than 1 user).

### DIFF
--- a/plugins/plugin-site/src/components/Plugin.jsx
+++ b/plugins/plugin-site/src/components/Plugin.jsx
@@ -25,7 +25,7 @@ function Developers({developers}) {
 Developers.propTypes = PluginDevelopers.propTypes;
 
 function Plugin({plugin: {name, title, stats, labels, excerpt, developers, buildDate, releaseTimestamp, healthScore}}) {
-
+    const installStr = stats.currentInstallPercentage ? formatPercentage(stats.currentInstallPercentage) : '?';
     return (
         <div onClick={() => navigate(`/${name}/`)} className="Plugin--PluginContainer">
             <div className="Plugin--IconContainer">
@@ -35,8 +35,7 @@ function Plugin({plugin: {name, title, stats, labels, excerpt, developers, build
                 <h4>{cleanTitle(title)}</h4>
             </div>
             <div className="Plugin--InstallsContainer">
-                {'Installs:  '}
-                {stats.currentInstallPercentage ? formatPercentage(stats.currentInstallPercentage) : '?'}
+                {`Used by ${installStr} of instances`}
             </div>
             <div className="Plugin--VersionContainer">
                 <span className="jc">
--- a/plugins/plugin-site/src/components/Plugin.jsx
+++ b/plugins/plugin-site/src/components/Plugin.jsx
@@ -36,7 +36,7 @@ function Plugin({plugin: {name, title, stats, labels, excerpt, developers, build
             </div>
             <div className="Plugin--InstallsContainer">
                 {'Installs:  '}
-                {formatPercentage(stats.currentInstallPercentage)}
+                {stats.currentInstallPercentage ? formatPercentage(stats.currentInstallPercentage) : '?'}
             </div>
             <div className="Plugin--VersionContainer">
                 <span className="jc">
--- a/plugins/plugin-site/src/components/Plugin.jsx
+++ b/plugins/plugin-site/src/components/Plugin.jsx
@@ -1,7 +1,7 @@
 import PropTypes from 'prop-types';
 import React from 'react';
 import {navigate} from 'gatsby';
-import {cleanTitle} from '../commons/helper';
+import {cleanTitle, formatPercentage} from '../commons/helper';
 import Icon from '../components/Icon';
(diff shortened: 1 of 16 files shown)


plugins/plugin-site/src/components/Plugin.jsx
```diff
@@ -36,7 +36,7 @@ function Plugin({plugin: {name, title, stats, labels, excerpt, developers, build
             </div>
             <div className="Plugin--InstallsContainer">
                 {'Installs:  '}
-                {stats.currentInstalls}
+                {stats.currentInstallPercentage ? formatPercentage(stats.currentInstallPercentage) : '?'}
```
daniel-beck: When there are no usage stats, it's 0 (or 0%).

We could mention that for new plugins, it's expected, but there are a few plugins hovering around 1-2 installs, and for them, 0 means 0, not "too new".

plugins/plugin-site/src/components/Plugin.jsx
```diff
@@ -36,7 +36,7 @@ function Plugin({plugin: {name, title, stats, labels, excerpt, developers, build
             </div>
             <div className="Plugin--InstallsContainer">
                 {'Installs:  '}
```
daniel-beck: Does this need a new label, e.g. "Popularity" or "Install Stats"? Or just "Installed on " / "Installed by"? Is that a complete sentence?

daniel-beck (approved):
Thanks for taking care of this!

I think this is a nice improvement as it's more readily understandable what the number means without having to figure out what 100% is, and it also decouples the per-plugin stats from general reporting behavior if the total number of reporting instances fluctuates.

plugins/plugin-site/src/components/Plugin.jsx
zbynek: With no usage stats we already store 0. This is just to avoid failing when the site expects new format and Algolia has old one (happens on PR preview site + maybe temporarily after this is merged and Algolia is updating...).

plugins/plugin-site/src/components/Plugin.jsx
zbynek: Good point. In the search page we could squeze in "installed on X% instances"
image
I'll check if we can do that somehow also for plugin pages.

daniel-beck: Grammar-wise, I think it would be "of instances"? Also note that the string is longer when it's very few installs. [1 install](https://deploy-preview-1499--jenkins-plugin-site-pr.netlify.app/kobiton-integration/) is `0.00033%`.

zbynek: Would "Install rate" make sense?  @MarkEWaite thoughts?

MarkEWaite: > Would "Install rate" make sense? @MarkEWaite thoughts?

I'm accustomed to "rate" having a stronger relationship to time than this measure seems to have.  

I like "Installed on 35% of instances" as much clearer than "105000".

zbynek: So I changed it to "used by" in search screen where there is less space and "installed on" in the plugin page where a line break is possible. If anyone has a preference for some other wording I can change it.

daniel-beck: > Screenshot 2023-09-22 at 14 34 46

Fits nicely (fake data) 👍

PR #1556, opened 2023-11-01
Add dark mode

### LINKED ISSUE #1146: Add a darkmode
Currently, https://plugins.jenkins.io/ has no dark mode at all. It would be pretty awesome, if we would have a darkmode switch on the page, allowing the user to toggle the theme or rely on the system theme and switch automatically.

zbynek:
Fixes #1146 

Before merging the header component should be updated.

### DIFF
--- a/plugins/plugin-site/gatsby-config.mjs
+++ b/plugins/plugin-site/gatsby-config.mjs
@@ -1,5 +1,6 @@
 import dotenv from 'dotenv';
 import {dirname} from 'path';
+import fs from 'fs';
 import {fileURLToPath} from 'url';
 import algoliaQueries from './src/utils/algolia-queries.mjs';
 
@@ -28,6 +29,9 @@ Object.keys(process.env).forEach(key => {
     element[splits.slice(-1)[0]] = process.env[key];
 });
 
+const darkCodeTheme = fs.readFileSync('../../node_modules/github-syntax-dark/lib/github-dark.css', 'utf8');
+const autoCodeTheme = `@media (prefers-color-scheme: dark){${darkCodeTheme}}`;
+fs.writeFileSync('../../node_modules/github-syntax-dark/lib/github-auto.css', autoCodeTheme);
 // This is the content of your gatsby-config.js
 // and what you need to provide as schema:
 const config = {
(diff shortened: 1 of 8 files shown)


plugins/plugin-site/gatsby-config.mjs
```diff
@@ -28,6 +29,9 @@ Object.keys(process.env).forEach(key => {
     element[splits.slice(-1)[0]] = process.env[key];
 });
 
+const darkCodeTheme = fs.readFileSync('../../node_modules/github-syntax-dark/lib/github-dark.css', 'utf8');
```
halkeye: can this be done with require.resolve? or import.meta.resolve? https://stackoverflow.com/questions/54977743/do-require-resolve-for-es-modules

seeing node_modules in paths always make me nervous, and .../.../node_modules doubly so.

plugins/plugin-site/src/styles/base.css
```diff
@@ -992,3 +995,65 @@ a.card:hover {
...
+    --bs-highlight-color: #dee2e6;
+    --bs-highlight-bg: #664d03;
+    --bs-border-color: #495057;
+    --bs-border-color-translucent: rgba(255, 255, 255, 0.15);
+    --bs-form-valid-color: #75b798;
+    --bs-form-valid-border-color: #75b798;
+    --bs-form-invalid-color: #ea868f;
+    --bs-form-invalid-border-color: #ea868f;
+  }
+
```
halkeye:
```suggestion
```

plugins/plugin-site/src/styles/base.css
```diff
@@ -992,3 +995,65 @@ a.card:hover {
     max-width: 700px;
   }
 }
+@media (prefers-color-scheme: dark) {
```
halkeye, line 994: how does this work for non dark? all the variables are already defined? or just empty?

zbynek: All of them are defined, either in jenkins.io stylesheet or Bootstrap. The bootstrap values are taken from Bootstrap's dark mode (with link color modified).

zbynek: All of them are defined, either in jenkins.io stylesheet or Bootstrap. The bootstrap values are taken from Bootstrap's dark mode (with link color modified).

timja (approved):
Looks good enough to me :) (tested in the deploy-preview), can be tweaked later.

Thanks for this!

NotMyFault:
I'm late to the party but thanks for the PR!

I've tried it locally and found no issues so far. Looks really good

PR #1573, opened 2023-12-21
add pagination to EOP

### LINKED ISSUE #1572: Enhancement: Add Pagination--Container at the bottom of the page
Add another Pagination--Container at the end of the search results. It's quite easy to navigate between pages at the bottom of site, otherwise we should scroll to the top to navigate between the pages

https://github.com/jenkins-infra/plugin-site/assets/107404972/21962db2-7f92-4e2b-8ce2-10ddbc3799ec

sridamul:
fix #1572 

Added Pagination--Container at the EOP for easy navigation.

### DIFF
--- a/plugins/plugin-site/src/components/Pages.jsx
+++ b/plugins/plugin-site/src/components/Pages.jsx
@@ -18,7 +18,7 @@ export default class Pages extends React.PureComponent {
 
         return (
             <ReactPaginate
-                pageCount={pages - 1}
+                pageCount={pages}
                 pageRangeDisplayed={this.props.pagesToDisplay}
                 marginPagesDisplayed={marginPagesDisplayed}
                 onPageChange={handlePageClick}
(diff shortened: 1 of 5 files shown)


sridamul:
@zbynek Can you review this?

zbynek:
I think it's enough to have the bottom pagination, the top one can be removed. Somewhat related suggestion: when there's only one page, there should be no pagination at all.

sridamul:
@zbynek Can you review this now?

sridamul:
@zbynek May I know why the checks failed?

sridamul:
@zbynek The Results seems to work only if result.pages > 2. I have tried with '1' and the page count is pages - 1.
So, Please verify the results.

sridamul:
result.pages > 1

pagination!

resutl.pages > 2

pagination2

works as expected when we use '2' instead of '1'
@zbynek Can you review this?

zbynek:
> works as expected when we use '2' instead of '1'

You're right that your code works, but I think we should fix the underlying inconsistency where number of pages is off by one. I've done that in a follow-up commit and merged this. Thanks for your contribution.

PR #1565, opened 2023-12-06
feat: unique Jenkinsfile

lemeurherve:
This PR merges the Jenkinsfile used on ci.jenkins.io with the Jenkinsfile_k8s used on infra.ci.jenkins.io

infra.ci.jenkins.io test build in #1569: https://infra.ci.jenkins.io/job/website-jobs/job/plugin-site/view/change-requests/job/PR-1569/1/console

Needs:
- https://github.com/jenkins-infra/kubernetes-management/pull/4750

Ref:
- Fixes #1377
- Closes #1569

PR #1575, opened 2023-12-22
chore: use exponential delay when axios retries requests

dduportal:
As mentionned by @slide and @uhafner in gitter/matrix channels, there has been issues with the https://plugins.jenkins.io/token-macro/ page not showing the proper readme page which is unexpected.

While searching for this, it appeared that there are a lot of network error on the node http.js client side like the following:

```text
➤ YN0000: [@jenkins-cd/jenkins-plugin-site]: error error trying to fetch https://plugins.jenkins.io/api/plugin/kubernetes-credentials-provider 
➤ YN0000: [@jenkins-cd/jenkins-plugin-site]: 
➤ YN0000: [@jenkins-cd/jenkins-plugin-site]: 
➤ YN0000: [@jenkins-cd/jenkins-plugin-site]:   AggregateError: 
➤ YN0000: [@jenkins-cd/jenkins-plugin-site]:   
➤ YN0000: [@jenkins-cd/jenkins-plugin-site]:   - AxiosError.js:89 Function.AxiosError.from
➤ YN0000: [@jenkins-cd/jenkins-plugin-site]:     [website-jobs_plugin-site_master]/[axios]/lib/core/AxiosError.js:89:14
➤ YN0000: [@jenkins-cd/jenkins-plugin-site]:   
➤ YN0000: [@jenkins-cd/jenkins-plugin-site]:   - http.js:606 RedirectableRequest.handleRequestError
➤ YN0000: [@jenkins-cd/jenkins-plugin-site]:     [website-jobs_plugin-site_master]/[axios]/lib/adapters/http.js:606:25
➤ YN0000: [@jenkins-cd/jenkins-plugin-site]:   
➤ YN0000: [@jenkins-cd/jenkins-plugin-site]:   - node:events:514 RedirectableRequest.emit
➤ YN0000: [@jenkins-cd/jenkins-plugin-site]:     node:events:514:28
➤ YN0000: [@jenkins-cd/jenkins-plugin-site]:   
➤ YN0000: [@jenkins-cd/jenkins-plugin-site]:   - index.js:14 ClientRequest.eventHandlers.<computed>
➤ YN0000: [@jenkins-cd/jenkins-plugin-site]:     [website-jobs_plugin-site_master]/[follow-redirects]/index.js:14:24
➤ YN0000: [@jenkins-cd/jenkins-plugin-site]:   
➤ YN0000: [@jenkins-cd/jenkins-plugin-site]:   - node:events:514 ClientRequest.emit
[...]

PR #1576, opened 2023-12-22
chore: add `retry(5)` to the Jenkinsfile

lemeurherve:
I removed it in #1565, putting it back trying to get builds passing.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -4,6 +4,7 @@ pipeline {
     ansiColor('xterm')
     disableConcurrentBuilds(abortPrevious: true)
     buildDiscarder logRotator(artifactDaysToKeepStr: '', artifactNumToKeepStr: '', daysToKeepStr: '5', numToKeepStr: '5')
+    retry(5)
   }
 
   agent {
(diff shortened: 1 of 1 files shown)


zbynek:
LGTM, but why is the build failing?

dduportal:
> LGTM, but why is the build failing?

Not sure why to be honest :| There have been error when the build fetches the plugin contents since months. There used to be HTTP/500 from the backend which has been fixed 2 months ago (OOM in the JVM).

Now we see TCP/TLS errors like the one described in https://github.com/jenkins-infra/plugin-site/pull/1575#issue-2053569162  but only in infra.ci.jenkins.io, not on ci.jenkins.io which is weird.

dduportal:
> I removed it in #1565, putting it back trying to get builds passing.

Problem is that the initial `Build` stage fails, so infra.ci.jenkins.io never reaches the "Deploy" phase (which is retried 5 times), so the build fails without retrying alas.

lemeurherve:
Superseded by https://github.com/jenkins-infra/plugin-site/pull/1575

zbynek:
But... API failing for a single plugin should result in a single docs page missing, not in whole build failing.

dduportal:
> But... API failing for a single plugin should result in a single docs page missing, not in whole build failing.

I wholeheartedly agree with you! But i have no idea how to do it with this toolchain (out of my depth here)

Issue #1580, opened 2023-12-30
Feature: Need proper formatting of the information for health score


AayushSaini101:
The description of the probes needs to be properly formatted, We can add the proper use of the list to show them
<img width="1279" alt="image" src="https://github.com/jenkins-infra/plugin-site/assets/60972989/2655b81a-11ee-48da-8268-a2bdf2cc0f4e">

PR #1581, opened 2023-12-30
Update description formatting

AayushSaini101:
Related: https://github.com/jenkins-infra/plugin-site/issues/1580


zbynek:
Thanks for the contribution. 

As far as design goes, personally I'm not sure if this is an improvement 🤷 @alecharp do you have an opinion?

From the technical point of view, there should be an `ul` tag wrapping the `li`s.

alecharp:
I agree with @zbynek this doesn't look right. Have you tested your changes locally @AayushSaini101?

alecharp:
I think we should close this for now.

Issue #1555, opened 2023-10-26
idea: use color in health check rows


halkeye:
Either whole row or just partial. Bootstrap has success, warn and danger colors. Could even color $percent of the background to make it easy to compare.

A white 0 next to a white 100 isn't super easy to judge

alecharp:
I like that idea. 
I tried to used colors on the initial pull requests but my taste for color isn't great. 

I would suggest not to use `error` (or `danger` or equivalent) to anything that is not `100/100`.
Not getting a perfect score is not synonym to an error, but to a suggestion of improvement. 

For example, not having JEP-229 implemented is not a problem, a plugin can be well maintained without it. However, it could help the plugin be released more often and that would benefit the users. But it's not a problem not to have it.

zbynek:
Maybe 0 score for important probe can be danger , for unimportant (weight 0) can be info, higher scores can be warn and success. Also to support colourblind users it would be nice to add icons (checkmark, cross, ...) .

AayushSaini101:
@zbynek Can i work on this issue,

zbynek:
@AayushSaini101 sure. Please open a pull request when you want to get some feedback on your work, thanks.

Ratul-Saha-003:
Is this issue still open? Can I work on this?

zbynek:
Fixed in https://github.com/jenkins-infra/plugin-site/pull/1593

PR #1584, opened 2024-01-05
Change colors of health checker results

AayushSaini101:
Related: https://github.com/jenkins-infra/plugin-site/issues/1555
SuggestedColors and Range: 
image

### DIFF
--- a/plugins/plugin-site/src/components/PluginHealthScoreProgressBar.jsx
+++ b/plugins/plugin-site/src/components/PluginHealthScoreProgressBar.jsx
@@ -1,11 +1,13 @@
 import React, {useState} from 'react';
 import PropTypes from 'prop-types';
 import {Progress, Tooltip} from 'reactstrap';
+ import '../styles/progressbar.css';
 
 function PluginHealthScoreProgressBar({healthScore, name}) {
     const score = healthScore.value || 0;
     const color =
-    score > 80 ? 'success' : score > 60 ? 'warning' : 'danger';
+    score >= 80 ? "first" : score >=60 && score <80 ? 'second' : score >=40 && score <60 ? 'third'
+     : score >=20 && score <40 ? 'fourth': 'fifth' ;
 
     const tooltipId = `tooltip-${name}`;
     const [tooltipOpen, setTooltipOpen] = useState(false);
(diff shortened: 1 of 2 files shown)


plugins/plugin-site/src/components/PluginHealthScoreProgressBar.jsx
```diff
@@ -1,11 +1,13 @@
 import React, {useState} from 'react';
 import PropTypes from 'prop-types';
 import {Progress, Tooltip} from 'reactstrap';
+ import '../styles/progressbar.css';
```
halkeye, line 4: can you change this to a css module, so we don't introduce global styles with generic names?

plugins/plugin-site/src/components/PluginHealthScoreProgressBar.jsx
```diff
@@ -1,11 +1,13 @@
...
 
 function PluginHealthScoreProgressBar({healthScore, name}) {
     const score = healthScore.value || 0;
     const color =
-    score > 80 ? 'success' : score > 60 ? 'warning' : 'danger';
+    score >= 80 ? "first" : score >=60 && score <80 ? 'second' : score >=40 && score <60 ? 'third'
```
halkeye, line 9: I'm not sure this needs to be css classes for a single style attribute change.

halkeye:
@alecharp

zbynek:
The referenced issue is about the health-score tab, which is not affected by this PR. Instead this changes the color of helth bars, which are already using reasonable colors. The current scheme seems logical (resembles traffic lights), the proposed one seems random. 

@AayushSaini101 I appreciate the effort, but I currently don't see this as an improvement. Contributions that would make the health-score tab (e.g. https://plugins.jenkins.io/slack/healthscore/ ) more colorful would be appreciated though!

@halkeye @alecharp I suggest to close this.

alecharp:
I agree we should close this.
I have some issues with hard coded values for the thresholds like that. 

There is a `statistic` object in the response the project is getting from Plugin Health Scoring could be used to have the threshold depending on the statistical scores across all the plugins (q1,q3, medium).

PR #1593, opened 2024-02-16
Improve plugin health score details visibility

alecharp:
Here I propose to collapse / show score section when necessary.
By default, when a section is valid, we don't see to all its details. This focus the eyes on the problematic spots.

I also added icons to easily see what section / detail need to be worked on.

<details>
<summary>Before</summary>

Screenshot 2024-02-16 at 14 16 54

</details>

<details>
<summary>After</summary>

Screenshot 2024-02-16 at 14 17 05

</details>

### DIFF
--- a/plugins/plugin-site/src/components/PluginHealthScore.css
+++ b/plugins/plugin-site/src/components/PluginHealthScore.css
@@ -34,7 +34,7 @@
     display: flex;
     justify-content: flex-end;
 }
-.pluginHealth--score-section--collapsible-icon img {
+.pluginHealth--score-section--collapsible-icon ion-icon {
     width: 14px;
 }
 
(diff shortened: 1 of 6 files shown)


MarkEWaite:
Those screenshots look great.  Nice use of the icon and color on the heading line to show me which areas need my attention.  Green is good.  Thanks!

alecharp:
The color are coming from the colors in the Jenkins Design Library. Because it is to change a SVG, I had to use the CSS `filter` feature.

NotMyFault (approved):
Really helpful

zbynek:
I'd suggest to use the `<ion-icon>` component for the images directly, that way the image is inlined and color can be changed by simply setting `color` in CSS (no need to add SVGs to repository). It would help with the chevron icons in dark mode, currently they are black instead of text color.

`<ion-icon name="chevron-down-outline"></ion-icon>`
`<ion-icon name="chevron-up-outline"></ion-icon>`

alecharp:
I didn't catch it was available in the dependencies! I'll make sure to change to that shortly.

alecharp:
I used your recommendation @zbynek and the preview is available to check the result. Let me know if there is anything else I can improve.

MarkEWaite:
I've confirmed that the preview site looks good.  Thanks!

MarkEWaite (approved):

Issue #1577, opened 2023-12-22
Plugin card view is difficult to read with dark mode enabled


NotMyFault:
When enabling dark mode, the plugin search card mode is difficult to read, given the text and health score uses almost the similar gray tone like the background:

<img width="1088" alt="Screenshot 2023-12-22 at 21 59 13" src="https://github.com/jenkins-infra/plugin-site/assets/13383509/01d633a6-656b-4215-a9ee-55a82902d16d">

nouralmulhem:
can you assign this on me ? check this out

## **Before:**
![image](https://github.com/jenkins-infra/plugin-site/assets/76218033/0aae72fe-6044-42f8-ab42-8d521fd5d618)

## **After:**
![image](https://github.com/jenkins-infra/plugin-site/assets/76218033/6bbdefd9-8fa6-404a-9c44-edeca43ce6b4)

NotMyFault:
Fixed in https://github.com/jenkins-infra/plugin-site/pull/1600

PR #1600, opened 2024-02-29
fix: update plugin card color on dark theme

nouralmulhem:
related issue: #1577 

## **Before:**

image

## **After:**

image

### DIFF
--- a/plugins/plugin-site/src/components/SearchResults.css
+++ b/plugins/plugin-site/src/components/SearchResults.css
@@ -47,7 +47,7 @@
     position: relative;
     display: block;
     & * {
-        color: #666;
+        color: var(--color);
     }
     & a, & a:visited {
         color: var(--link-color);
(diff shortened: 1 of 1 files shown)


zbynek:
Thanks! Definitely looks better in dark mode. 

This also changes the color in light mode, causing the "icons" (`.Plugin--IconContainer i`) to have too low contrast now. Since the background for those is the same in light/dark mode, they should also use the same foreground in both.

nouralmulhem:
thank you for noticing this!! , I think this has nothing to do with the fact that the text color has to change, wanted also to refer to which was presented previously

image

as you said i recommend making **`icon text foreground color`** unified (not the whole text), and this should be the solution, as the background color doesn't change

i will update my branch with the proposed solution and maybe you can share your thoughts again with me

zbynek (approved):
thanks!

nouralmulhem:
@zbynek I'm kind of new, should I close and merge?

NotMyFault (approved):
Thanks!

PR #1626, opened 2024-03-02
Update renovate schedule

NotMyFault:
The change proposed updates the renovate schedule to run more frequent than on the first date of the month.
This repository has numerous dependencies and the first of the month doesn't suffice to tackle the entire log, causing a backlog becoming longer per month, I occasionally trigger by hand to empty the queue.

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -1,7 +1,6 @@
 {
     "extends": [
-        "config:base",
-        "schedule:monthly"
+        "config:base"
     ],
     "labels": [
         "dependencies"


lemeurherve (approved):

zbynek:
With dependabot we had weekly updates and I changed them to monthly, because more frequent updates just meant updating minor versions of the same library every week even though it has no impact on the site. It would still be useful to have some kind of throttling to make sure a particular dependency does not get updated too often. Maybe keep monthly updates but increase the max number of parallel PRs?

PR #1685, opened 2024-04-05
Uses static file from reports for the scores

alecharp:
Related to jenkins-infra/helpdesk#3961.


zbynek:
Thanks for the PR. The mock request(s?) in tests also need updating.

alecharp:
Sorry, I ran `yarn test` before pushing but I might have been in the site folder, not at the root level. 
I'll fix that shortly

alecharp:
Sorry I couldn't correct the PR before you did. This pull request was not urgent to me as using the API would still work.

PR #1693, opened 2024-04-07
Display message when no open issues

### LINKED ISSUE #1691: Outline if a plugin has no open issues
If a plugin has no open issues, I would recommend displaying that somewhere on the `/issues/` tab.
Currently, you get the same view, like if the table isn't loading, if a plugin has no open issues, which has tricked me quite a few times before ;) 

Screenshot 2024-04-07 at 12 08 40

sridamul:
Closes #1691 

Display a message when there are no open issues instead of showing an empty table.
Eg: [Console Column plugin](https://plugins.jenkins.io/console-column-plugin/issues/)

isssue msg

### DIFF
--- a/plugins/plugin-site/src/components/PluginIssues.jsx
+++ b/plugins/plugin-site/src/components/PluginIssues.jsx
@@ -44,7 +44,7 @@ function PluginIssues({pluginId}) {
                             </tr>
                         </thead>
                         <tbody>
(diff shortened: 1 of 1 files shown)


zbynek:
The old code also checked that issues is truthy,  I'd suggest to add that  check to the new code as well. Otherwise 👍

PR #1867, opened 2024-08-24
Use dark theme variables from Jenkins.io

janfaracik:
Jenkins.io's stylesheet now includes variables for dark theme, this PR updates the plugin site to use them for consistency. Also updates the 'How to install' button to use the button class from Jenkins.io.

##### Before
image

##### After
image

### DIFF
--- a/plugins/plugin-site/gatsby-config.mjs
+++ b/plugins/plugin-site/gatsby-config.mjs
@@ -61,8 +61,7 @@ config.plugins = [
                 '@import \'./styles/base.css\';',
                 '@import \'./styles/font-icons.css\';',
                 '@import \'github-syntax-light/lib/github-light.css\';',
-                '@import \'github-syntax-dark/lib/github-auto.css\';',
-                '@import \'https://www.jenkins.io/stylesheets/styles.css\';'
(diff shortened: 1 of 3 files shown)


plugins/plugin-site/gatsby-config.mjs
```diff
@@ -61,7 +61,8 @@ config.plugins = [
                 '@import \'./styles/base.css\';',
                 '@import \'./styles/font-icons.css\';',
                 '@import \'github-syntax-light/lib/github-light.css\';',
-                '@import \'github-syntax-dark/lib/github-auto.css\';'
+                '@import \'github-syntax-dark/lib/github-auto.css\';',
+                '@import \'https://www.jenkins.io/stylesheets/styles.css\';'
```
zbynek: This stylesheet should be loaded by default, doesn't have to be part of extra CSS

zbynek (changes requested):

plugins/plugin-site/gatsby-config.mjs
janfaracik: Ah fantastic - I hadn't realised that.

PR #1926, opened 2024-10-05
Move rehype plugins into this repo

zbynek:
Inlines the code from archived repository: https://github.com/halkeye/gatsby-plugins
This allows some dependencies like rehype to be updated.


zbynek:
@dduportal any idea why this fails on infra.ci and passes on ci?

PR #2055, opened 2025-01-03
feat(community): add thank-a-contributor widget

krisstern:
As discussed previously during Jenkins Advocacy and Outreach SIG meetings, adding the thank-a-contributor widget to the bottom of the plugins site landing page just above the official footer. 

c.c. @MarkEWaite @alyssat @gounthar. 

Look of feature in dark mode:
Screenshot 2025-01-04 at 4 07 50 AM


gounthar (approved):
Thanks, Kris! 👌

plugins/plugin-site/src/components/ThankAContributorNote.jsx
```diff
@@ -0,0 +1,150 @@
...
+        axios
+            .get(
+                'https://raw.githubusercontent.com/jenkins-infra/jenkins-contribution-stats/main/data/honored_contributor.csv',
+                {responseType: 'text'}
+            )
+            .then((response) => {
+                setThankYou(Papa.parse(response.data)?.data[1]);
+            });
+
+        const interval = setInterval(() => {
```
zbynek, line 41: This is to make sure the contributor is refreshed after 1 hour of having the page open? I don't think that's needed.

plugins/plugin-site/src/components/ThankAContributorNote.jsx
```diff
@@ -0,0 +1,150 @@
...
+            sx={{
+                padding: theme.spacing(5, 2),
+                display: 'flex',
+                justifyContent: 'center',
+                alignItems: 'center',
+            }}
+        >
+            <Box
+                sx={{
+                    padding: isMobile ? theme.spacing(2) : theme.spacing(3),
```
zbynek: It would be more consistent to extract these into a CSS file, then the media queries can be done in CSS, jsx would only set class names.

plugins/plugin-site/package.json
```diff
@@ -66,11 +68,13 @@
     "@jenkins-cd/gatsby-source-jenkinsplugins": "*",
     "@jenkins-cd/gatsby-transformer-rehype": "*",
     "@juggle/resize-observer": "3.4.0",
+    "@mui/material": "^6.3.1",
```
zbynek: I don't think we need a big UI package (https://packagephobia.com/result?p=%40mui%2Fmaterial) just to style a few `div`s, especially since we're overriding a lot of those styles anyway.

plugins/plugin-site/package.json
```diff
@@ -58,6 +58,8 @@
   "homepage": "https://plugins.jenkins.io/",
   "dependencies": {
     "@babel/core": "7.26.0",
+    "@emotion/react": "^11.14.0",
+    "@emotion/styled": "^11.14.0",
```
zbynek: Are these two used anywhere?

plugins/plugin-site/src/components/ThankAContributorNote.jsx
```diff
@@ -0,0 +1,150 @@
...
+    const [thankYou, setThankYou] = useState([]);
+
+    useEffect(() => {
+        axios
+            .get(
+                'https://raw.githubusercontent.com/jenkins-infra/jenkins-contribution-stats/main/data/honored_contributor.csv',
+                {responseType: 'text'}
+            )
+            .then((response) => {
+                setThankYou(Papa.parse(response.data)?.data[1]);
```
zbynek: For readability it would be better to convert the array into an object, e.g. `{name: "Alice", avatar: "https://.....png", repos: ["git-plugin", "svn-plugin"]... }` -- it would also avoid calling `split` and `replace` repeatedly.

plugins/plugin-site/package.json
krisstern: Yes, for the componenets of the component library MUI in new feature

plugins/plugin-site/package.json
krisstern: Removed

plugins/plugin-site/package.json
krisstern: Removed with component library

plugins/plugin-site/src/components/ThankAContributorNote.jsx
krisstern: Done

plugins/plugin-site/src/components/ThankAContributorNote.jsx
krisstern: It is more a failsafe mechanism, in case the page is left open for a very long time > an hour

plugins/plugin-site/src/components/ThankAContributorNote.jsx
krisstern: Still need to call split and replace repeatedly though

plugins/plugin-site/src/components/ThankAContributorNote.jsx
krisstern: Done

PR #2125, opened 2025-03-01
Update thank you note in the Plugin page

### LINKED ISSUE #2124: Standardize Thank You Note on Plugins Page to Match Root & Downloads Page
The Thank You note at the end of the Plugins page on [Jenkins.io](https://www.jenkins.io/) is different from the ones on the Root page and Downloads page. To ensure consistency, it should be updated to match the format used on the other pages.

I want to work on this issue.

lakshmishreea122003:
Fixes https://github.com/jenkins-infra/plugin-site/issues/2124

**With the updates it looks like**
Screenshot 2025-03-01 235641

This is similar to the thank you note in the Root page and the Downloads page


zbynek (changes requested):
Let's use the same class names as the download page, then we can just delete the whole CSS file here.

lakshmishreea122003:
> Let's use the same class names as the download page, then we can just delete the whole CSS file here.

@zbynek Sorry for the late reply I was in a hackathon. I did update the class names to be similar to the names in the downloads page and I deleted the CSS file. Let me know if any more changes required.

plugins/plugin-site/src/components/ThankAContributorNote.jsx
```diff
@@ -70,80 +65,77 @@ function ThankAContributorNote() {
...
+                        className="image-div"
                     >
                         <img
                             src={imageError || !thankYou['GH_HANDLE_AVATAR'] ? placeholderImage : thankYou['GH_HANDLE_AVATAR'].replace(/['"]+/g, '')}
                             alt="Random contributor image"
-                            width={isDesktop ? 100 : isMobile ? 36 : 90}
-                            height={
-                                isDesktop ? 100 : isMobile ? '100%' : 90
-                            }
-                            className="thank-you-note-img"
```
krisstern, line 91: This would get rid of the custom responsiveness behavior. Have you checked if it would still work without this?

plugins/plugin-site/src/components/ThankAContributorNote.jsx
```diff
@@ -4,14 +4,9 @@ import React, {useEffect, useState} from 'react';
...
 import dayjs from 'dayjs';
 import Papa from 'papaparse';
-import {useMediaQuery} from '../hooks/useMediaQuery';
-import './ThankAContributorNote.css';
 import placeholderImage from '../images/person-circle-outline.svg';
 
 function ThankAContributorNote() {
-    const isMobile = useMediaQuery('(max-width: 768px)');
-    const isDesktop = useMediaQuery('(min-width: 1024px)');
-
```
krisstern, line 14: This would get rid of the custom responsiveness behavior. Have you checked if it would still work without this?

krisstern:
@lakshmishreea122003 Are you sure the new changes would work? I checked there is no `desc-div` class in this repo for example.

krisstern:
Also, have you checked if this repo still works with the current `package.json` file settings as it stands? I tried deleting all `node_modules` folder and started `npm i` again but ran into some issues. I am currently trying to see if there can be a quick fix.

plugins/plugin-site/src/components/ThankAContributorNote.jsx
krisstern: Please revert these changes, I have checked it no longer works as expected after you removed these. Also, please test carefully before pushing your code to GitHub next ime. It appears you had not tested these changes previously.

plugins/plugin-site/src/components/ThankAContributorNote.jsx
krisstern: Please revert these changes, I have checked it no longer works as expected after you removed these. Also, please test carefully before pushing your code to GitHub next ime. It appears you had not tested these changes previously.

plugins/plugin-site/src/components/ThankAContributorNote.jsx
zbynek: If we want different size of the image for mobile/desktop devices, it should be defined in jenkins.io CSS, there should be no need to set these in `jsx`.

krisstern: Yes, agreed. But that is not currently done.

krisstern:
> @lakshmishreea122003 Are you sure the new changes would work? I checked there is no desc-div class in this repo for example.

I checked and don't think your code with the `desc-div` class is working currently

plugins/plugin-site/src/components/ThankAContributorNote.jsx
krisstern: It is okay to refactor the code, but it is not okay to change the design

lakshmishreea122003:
@krisstern I shall make the necessary changes as you have instructed.

For testing, I use Gitpod to build and test the UI. The video I shared is the site built and tested on Gitpod.

https://github.com/user-attachments/assets/48910179-d3a7-444b-9e09-244fe3487fb4

lakshmishreea122003:
I will also update the CSS in jenkins.io to ensure that the thank you note on the root page and the downloads page match the styling used on the Plugins site.

plugins/plugin-site/src/components/ThankAContributorNote.jsx
lakshmishreea122003: > This would get rid of the custom responsiveness behavior. Have you checked if it would still work without this?

I removed this because the jenkins.css file in jenkins.io already handles the responsive styling directly using media queries:

> @media (max-width: 768px) {
  .image-div img {
    height: 36px;
    width: 36px;
  }
  .contributor-data {
    flex-direction: column; 
    align-items: center;    
  }
  .image-div {
    margin-bottom: 0.5rem;  
  }
}
/* for screen sizes more than 768px */ 
> .image-div img {
  object-fit: contain;
  height: 90px;
  width: 90px;
  max-width: 100%;
}

Since the CSS already covers both mobile and desktop cases, I felt the inline isMobile/isDesktop logic could be safely removed.

lakshmishreea122003:
> > @lakshmishreea122003 Are you sure the new changes would work? I checked there is no desc-div class in this repo for example.
> 
> I checked and don't think your code with the `desc-div` class is working currently

the desc-div caters to the text data shown in the thank you note. In the site that I built using Gitpod it appears the same as it is in the [jenkins.io](https://www.jenkins.io/). Can you please tell what exactly does not work in desc-div class

krisstern:
Hi @lakshmishreea122003 can you just follow my original instruction to you and my intention to keep the original design in both the Contributor Spotlight and here intact, and not to add new ideas to this?

krisstern:
> the desc-div caters to the text data shown in the thank you note. In the site that I built using Gitpod it appears the same as it is in the [jenkins.io](https://www.jenkins.io/). Can you please tell what exactly does not work in desc-div class

No worries I double checked everything is in place, except for the responsive behavior is not what I like right now. 

I want for mobile:

Screenshot 2025-03-04 at 12 11 41 AM

Please do not change this.

lakshmishreea122003:
> Hi @lakshmishreea122003 can you just follow my original instruction to you and my intention to keep the original design in both the Contributor Spotlight and here intact, and not to add new ideas to this?

I will surely adhere to your instructions.

lakshmishreea122003:
> > the desc-div caters to the text data shown in the thank you note. In the site that I built using Gitpod it appears the same as it is in the [jenkins.io](https://www.jenkins.io/). Can you please tell what exactly does not work in desc-div class
> 
> No worries I double checked everything is in place, except for the responsive behavior is not what I like right now.
> 
> I want for mobile:
> 
> Screenshot 2025-03-04 at 12 11 41 AM
> Please do not change this.

I shall update the jenkins.css file in jenkins.io to make sure the design is as you expect.

lakshmishreea122003:
@krisstern I have updated the jenkins.css file in jenkins.io in https://github.com/jenkins-infra/jenkins.io/pull/7936. Please check, and let me know if any changes are required.

lakshmishreea122003:
@krisstern Please let me know if I need to make any modifications to this PR.

plugins/plugin-site/src/components/ThankAContributorNote.jsx
krisstern: The responsiveness was not there when I made the comment. You did add it after the fact.

krisstern (approved):
Looks okay to me after checking on Chrome, Firefox, as well as Firefox.

krisstern:
@zbynek Please give this a final review. Thanks!

lakshmishreea122003:
@zbynek please can you review this PR.

Issue #2180, opened 2025-04-13
Misaligned Close Button in Filter Tags on Plugin Page


singghh:
### Problem with this page

https://plugins.jenkins.io/

### Expected behavior

The close (✖) icons for the filter tags in the plugin search UI are overlapping with the tag text, making them visually cluttered and potentially hard to click.

Before:

![Image](https://github.com/user-attachments/assets/11a1fc6b-bec4-45c2-aa47-945cb6e916a3)

### Actual behavior

![Image](https://github.com/user-attachments/assets/0de23f4f-2add-4aec-a5b3-2678f92f1790)

There should be some space between those selected filter options.

singghh:
The issue is solved at #2181

PR #2181, opened 2025-04-13
Added padding-right:0.5rem

singghh:
#2180

### DIFF
--- a/plugins/plugin-site/src/styles/base.css
+++ b/plugins/plugin-site/src/styles/base.css
@@ -513,7 +513,7 @@ body #grid-box .NoLabels legend {
   margin-right:.5rem;
 }
 
-##pluginPage #grid-box .sidebar .label-link a{
+#pluginPage #grid-box .sidebar .label-link a{
   display:block;
   line-height: 1.5rem;
   padding: 4px 0;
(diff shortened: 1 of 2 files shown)


zbynek:
Thanks!

zbynek:
@singghh one suggestion for next time: use the `Fixes` keyword so that the linked issue is closed automatically when PR is merged.

plugins/plugin-site/src/styles/base.css
```diff
@@ -513,7 +513,7 @@ body #grid-box .NoLabels legend {
   margin-right:.5rem;
 }
 
-##pluginPage #grid-box .sidebar .label-link a{
+#pluginPage #grid-box .sidebar .label-link a{
   display:block;
   line-height: 1.5rem;
   padding: 4px 0;
```
daniel-beck, line 519: While this seems to have just been a typo previously, it doesn't look like anyone noticed and this block should have been removed instead as unnecessary.

IMO the new additional line-height looks wrong, and also does not extend to the list of maintainers, so this makes the sidepanel look inconsistent.

> Screenshot 2025-04-14 at 09 59 08

singghh: Yeah, i noticed and fixed that
Should i see that issue what you raised.
Regarding line height i can see that too.

zbynek: Removed the extra block of CSS to restore the style in `main`.

daniel-beck: @zbynek Thanks!

FTR https://github.com/jenkins-infra/plugin-site/commit/2e3c4d0571236f042b78c7eff7d4ac7e0a77bb7c

PR #2315, opened 2025-07-27
Remove matomo

zbynek:
The Matomo integration fails because of a wrong certificate and I'm not sure whether/how it was used while it worked. 

@halkeye @MarkEWaite ok to remove it?

Added in https://github.com/jenkins-infra/plugin-site/pull/755


MarkEWaite (approved):
Ok with me that we remove it.

halkeye:
I'm happy to finally delete that droplet
but i can also fix the cert pretty easy, i'm sure the script just got stuck

Issue #2352, opened 2025-08-22
Use links for cards in search results


daniel-beck:
### Problem with this page

https://plugins.jenkins.io/ui/search/?query=whatever

### Expected behavior

I want to open multiple search results in new browser tabs.

### Actual behavior

I cannot because there aren't actual links.

The best I can do is click the "?" to reach the health scores, and then switch "tabs" on the UI back to the main page.

soujanyabhirade:
Hi,

I am exploring Jenkins contributions for GSoC and would like to work on this issue if it is available.

Please let me know if it is okay for me to take this.

Thanks!

PR #2369, opened 2025-09-07
search result links and cleans up the unused variables in layout.jsx

adicodeswell:
Issue: #2352 

PR Title: feat: Allow search results to open in new tab and fix lint warnings

This PR fixes the search result links and cleans up the unused variables in layout.jsx.

1. Enables Opening Search Results in a New Tab

Currently, the search result items on the plugin index page are rendered as div elements with an onClick handler. This prevents users from using standard browser functionality like middle-clicking or Ctrl+Click to open plugin pages in a new tab, which is frustrating for users who want to compare multiple plugins.

This change replaces the div with a standard HTML <a> tag, restoring the expected browser behavior and improving usability.

[Fix video.zip](https://github.com/user-attachments/files/22198674/Fix.video.zip)

2. Removes Unused Variables in layout.jsx
The layout.jsx component contained two unused variables (buildTime and siteUrl) that were triggering ESLint no-unused-vars warnings during development. This change removes the unused variables, cleaning up the code and resolving the linting errors.

### DIFF
--- a/plugins/plugin-site/src/components/Plugin.jsx
+++ b/plugins/plugin-site/src/components/Plugin.jsx
@@ -1,6 +1,5 @@
 import PropTypes from 'prop-types';
 import React from 'react';
-import {navigate} from 'gatsby';
 import {cleanTitle, formatPercentage} from '../commons/helper';
 import Icon from '../components/Icon';
 import PluginLabels from '../components/PluginLabels';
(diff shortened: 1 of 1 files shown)


zbynek:
As per https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/a#technical_summary it's not allowed to have `<a>` nested in other `<a>` elements, which caused  some warnings during development, that's why the top level link is a `<div>`.

We could do one/both of
* wrap the the plugin name in `<a>` to make that a link
* enhance the card click handler to check for Ctrl+Click and open the page in a new tab in that case

adicodeswell:
Currently, I have replaced the outer `<div>` with an `<a>` tag. Is there anything wrong with this?

PR #2524, opened 2026-04-09
Fix formatting issues in index.jsx

pathum2583-eng:


zbynek:
@pathum2583-eng maybe check if your account was hacked.

PR #2563, opened 2026-07-06
chore(Jenkinsfile/CI) utilizes container agents instead of VM agents

dduportal:
Following https://github.com/jenkins-infra/helpdesk/issues/5202#issuecomment-4887424379, let's use container agents on ci.jenkins.io.


dduportal:
Self merging: CI checks are GitHub HTTP/429 (unrelated)

